PR #308, opened 2015-03-22
Protect all words (in titles) that should not undergo case changes

nickbart1980:
Protect all words (in titles) that should not undergo case changes when used with "sentence-case" styles such as biblatex-apa.


plk:
I have merged this into the DEV branch, thanks.

PR #330, opened 2015-07-24
Augment bibheading definitions to include subsubbibliographies

jeffreywildman:
Hi, would you consider the addition of bibheadings one level deeper than currently supported?  It would allow one to call:

    \printbibliography[heading=subsubbibliography]
    \printbibliography[heading=subsubbibnumbered]
    \printbibliography[heading=subsubbibintoc]

Besides the commit in the pull request, I've written a set of tests for each document class already supported in biblatex.def.  I've uploaded these tests to [gist](https://gist.github.com/jeffreywildman/580d536acbb663b7c06e).

### DIFF
--- a/tex/latex/biblatex/biblatex.def
+++ b/tex/latex/biblatex/biblatex.def
@@ -1800,6 +1800,13 @@
     \addcontentsline{toc}{subsection}{#1}}
   \defbibheading{subbibnumbered}[\refname]{%
     \subsection{#1}}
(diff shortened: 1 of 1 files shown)


plk:
@aboruvka  - what's your view on this?

plk:
We will have to consider this later as we're in the middle of considerable internal changes at the moment. I doubt we would want to do this in the core anyway as it's never requested and we only implement in the core what has a wide use-case.

PR #351, opened 2015-11-06
Consistently title-case (English) titles

nickbart1980:


josephwright:
Have you checked these against the original literature? The form that should be recorded is 'as published'.

nickbart1980:
> The form that should be recorded is 'as published'.

I don’t think so. I am not aware of any style manual (Chicago, APA, …) that wants titles ‘as published’. They either want title case, or sentence case. And the only way to have this formatted properly in biblatex output is indeed entering all (English) titles in title case.

plk:
I agree that this is really a style issue. I know that APA style mandates certain casing for certain. However, this is just a sample file and I have a lot of tests which are diffing PDFs and I don't to update all the reference data just for a sample file ...

PR #398, opened 2016-03-27
Catalan translation maintenance

sebasvila:
A new batch of enhancements of the catalan localization. Mostly, fine tunning of abbreviations:
- Rewrite month abbrev to standard form
- Unify some abbrev
- Rewrite some language abbrev's to current use
- Review vs revise
- Improve abbrev plural forms after a mistake: catalan prefix form abbrev's are number invariant.


plk:
Many thanks.

PR #417, opened 2016-05-03
Slovak support

DavidLuptak:
Shouldn't be at least 'langslovak' and 'fromslovak' included in \abx@dostrings to get it working?


plk:
Thanks - this was to be done when the doc was also updated but it's done now.

PR #423, opened 2016-05-19
Bumped control file version to 3.2

retrography:
Biber 2.6 (dev) already asks for it


plk:
This should already be done in the version uploaded yesterday. If you are using Git, I have only just now pushed those changes.

PR #426, opened 2016-05-21
Estonian

bkmgit:
Started template for Estonian. Would be helpful to have it added so native speakers can update it appropriately.


plk:
Thank you, I have added the other support necessary for this and it will be in the 3.5 DEV version shortly.

PR #434, opened 2016-06-06
update german.lbx

jspitz:
There is one open issue. I don't know how to translate `translationfrom`. The docs say: 
> The expression ‘translated from [the] language'.

But which bibstring is "[the] language" taken from? If it is `lang<lang>`, the inflexion will not be correct in German, since `lang<lang>` is nominative, whereas the phrase would require dative case (e.g. "übersetzt aus dem Englischen", not "übersetzt aus dem Englisch"). If it is `fromlang<lang>`, the case is correct, but looking at the other translations, I doubt this is used (due to the preposition used in `translationfrom`).


plk:
I would have thought that the `fromlanguage` was intended for the darive case - we'd have to check the styles.

jspitz:
Yes, `fromlanguage` uses the dative case, but it also uses prepositions. That's why I thought it is not used by the case where `translationfrom` is used, because this introduces its own prepositions. That is, in English, `translationfrom` is "translated from", `fromgerman` is "from the German". That would make "translated from from the German".

plk:
I'm not entirely sure the language after "translatedfrom" is supposed to come from a bibstring for exactly this reason. Howerver, @moewew - how do you deal with this?

moewew:
Mhhh, I could find no style that actually uses `translationfrom` and no other precedence either. In the `dutch.lbx` we can find `vertaald van` and `van` is a very unusual proposition after "vertaald" (at least I would say so, I'm not a native speaker, though). I don't speak French, but there the file says "traduit de" does that make sense? 

The English version "translation from" is also a bit inconsistent, either it is paired with `from<lang>` in which case there are too many prepositions (so that is not a good idea), or with `lang<lang>` in which case we don't get the "the" we get in the other cases.

jspitz:
Do we know the context where this bibstring has been introduced?

moewew:
Good question, a bit of digging reveals it was introduced in `biblatex` 2.0 with https://github.com/plk/biblatex/commit/c52a3e7108ff48da54b3c5cc93abdf55a8dbb234. Apparently for the `related` function - presumably to introduce a related entry. In that case a language would not follow automatically, that would have to be manually implemented by the style. But then I don't understand the use case for `translationfrom` since we already have `translationof` and `translatedas`. (I only just noted the discrepancy between `translationfrom` and its value "translated from" (noun vs. participle), something that threw me in the French version.)

Maybe @aboruvka could shed some more light on the intended use for the string?

PR #509, opened 2016-11-09
Remove monospace style from doi and eprint

tobiasdiez:
The doi and the eprint identifier were set in monospace font, which often results in an uneven display. Monospace is often used to increase readabilty (for example for passwords), but as most identifiers are as readable as ISBN/ISSN numbers a consistent font should be used for both.

### DIFF
--- a/tex/latex/biblatex/biblatex.def
+++ b/tex/latex/biblatex/biblatex.def
@@ -438,8 +438,8 @@
 \DeclareFieldFormat{doi}{%
   \mkbibacro{DOI}\addcolon\space
   \ifhyperref
-    {\href{https://doi.org/#1}{\nolinkurl{#1}}}
-    {\nolinkurl{#1}}}
+    {\href{https://doi.org/#1}{#1}}
+    {#1}}
 \DeclareFieldFormat{edition}{%
   \ifinteger{#1}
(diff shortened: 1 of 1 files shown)


johannesbottcher:
Instead of completely removing the switches, substituting them with a new macro defaulting to the current behaviour would be better. A user can redefine the font if needed. The PR in ist current state would change many many user documents.

moewew:
The final decision lies with PLK, but I absolutely agree with Johannes that this will have quite dramatic consequences on the output and backwards compatibility.
Plus, I for one actually like the monospaced look.
As the pull request shows it is easy enough for anyone to redefine the formats if need be (i.e. if they don't like the monospacing).

plk:
Yes, if it is changeable, I don't want to alter defaults.

josephwright:
In addition to the above, the entire point of `\nolinkurl` is that DOIs tend to be treated as 'URL stubs' so get formatted the same way. I'd be concerned about a change here.

tobiasdiez:
Ok, I understand and agree with your points. Would you accept a solution which extracts the formatting of the doi + eprint to a new command (which defaults to `\nolinkurl`) as proposed by @johannesbottcher ?
If yes, could you please point me to a "how to contribute guide" as I'm a bit lost where I should add the documentation part, which naming conventions you have etc.

PR #550, opened 2017-04-02
zap tabs

moewew:
replaced tabs by appropriate number of spaces


moewew:
@plk If you don't want to run tests against this (diffs confirms only whitespace changed, so there should be no problems), I will merge the changes shortly.

PR #552, opened 2017-04-03
Improve French translations

cmfwyp:
Changes:

- `\bibrangedash` defined as a hyphen instead of an en dash

  Rationale:

  - “L’abréviation de « pages » est « p. » (en anglais « pp. ») et on sépare les numéros par un tiret court « p. 12-34 » (en anglais par un moyen « pp. 12–34 »).” (https://jacques-andre.fr/faqtypo/lessons.pdf, page 43)

- `andothers`, `andmore` translated to “et coll.” instead of “et al.”

  Rationale:

  - “On peut aussi recourir à des formules telles que : *sous la direction de*, ou : *et collab*…” (http://www.orthotypographie.fr/volume-I/bandeau-bureau.html#Bibliographie)
  - “On tend à privilégier les formes françaises *et autres*, *et coll.*, etc., plutôt que la locution latine *et alii* ainsi que son abréviation *et al.*” (http://gdt.oqlf.gouv.qc.ca/ficheOqlf.aspx?Id_Fiche=8362481)
  - “S’il y a plus de trois auteurs, on indique le nom et le prénom du premier en les faisant suivre d’une virgule et de la mention *et autres* ou *et collab.* (pour *et collaborateurs*) de préférence à l’abréviation latine *et al.* (pour *et alii* « et les autres »).” (http://bdl.oqlf.gouv.qc.ca/bdl/gabarit_bdl.asp?t1=1&id=3245)

- `sections` and `sectiontotals` translated to “\S” instead of “\S\S”

  Rationale:

  - “On proscrira la répétition des signes ou des lettres pour indiquer la pluralité : § 5 et 6 et non §§ 5 et 6” (*Lexique des règles typographiques en usage à l’Imprimerie nationale* (1990), page 6)

- `annodomini` translated to “apr. J.-C.” instead of “ap. J.-C.”

  Rationale:

  - http://www.orthotypographie.fr/volume-I/abreviations-01.html
  - http://www.btb.termiumplus.gc.ca/redac-chap?lang=fra&lettr=chapsect1&info0=1.4#a
  - *Lexique des règles typographiques en usage à l’Imprimerie nationale* (1990), page 7.

- `opcit` translated to “ouvr. cité” instead of “op. cit.” and `loccit` translated to “pass. cité” instead of “loc. cit.”

  Rationale: http://www.orthotypographie.fr/volume-I/abreviations-01.html

- `sequens` and `sequentes` translated to “et suiv.” instead of “sq.” and “sqq.”

  Rationale: http://www.orthotypographie.fr/volume-I/abreviations-01.html

- `langestonian` translated to “estonien” and `fromestonian` translated to “de
  l’estonien”

- `langbulgarian` translated to “bulgare” and `frombulgarian` translated to “du
  bulgare”

### DIFF
--- a/tex/latex/biblatex/lbx/french.lbx
+++ b/tex/latex/biblatex/lbx/french.lbx
@@ -6,8 +6,7 @@
 \DeclareRedundantLanguages{french}{french}
 
 \DeclareBibliographyExtras{%
-  \protected\def\bibrangedash{%
-    \textendash\penalty\hyphenpenalty}% breakable dash
+  \protected\def\bibrangedash{-}
   \let\finalandcomma=\empty
   \let\finalandsemicolon=\empty
   \def\mkbibordinal{\mkbibmascord}%
@@ -113,7 +112,7 @@
   references       = {{R\'ef\'erences}{R\'ef\'erences}},
   shorthands       = {{Liste des sigles}{Sigles}},
   editor           = {{\'editeur}{\'ed\adddot}},
-  editors          = {{\'editeurs}{\'eds\adddot}},
+  editors          = {{\'editeurs}{\'ed\adddot}},
   compiler         = {{compilateur}{comp\adddot}},
   compilers        = {{compilateurs}{comp\adddot}},
   redactor         = {{r\'edacteur}{r\'ed\adddot}},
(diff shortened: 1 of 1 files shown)


dbitouze:
I second these relevant, well argued, changes. Just one point: please use commands for accents (e.g. `\'e` instead of `é`). And take this opportunity to fix the currently untranslated `circa`, `spring`, etc. :)

Evpok:
*et al.* and *op. cir.* are ubiquitous, even in French publication. The motivation is commendable, but changing those might lead to confusion from users. To me, it would feel awkward to read *et coll.* and I would not want to use it in my publications.

cmfwyp:
@dbitouze Done.

@Evpok Do you prefer *et collab.*? If not, I'll reluctantly revert to *et al.*

Evpok:
It would be less confusing, yes, but I would probably still override it to *et al.* in my documents. Sorry.

moewew:
Maybe you can submit a new pull request with the uncontroversial changes (`estonian`, `summer`, ...). You could then open another pull request with the more controversial ones and that could be discussed there.

If it turns out that Canadian French is significantly more hostile towards Latin abbreviations than continental French, we could think about splitting off `canadien.lbx` for Canadian French. (Haven't actually checked what that would entail, but it *should* work.) BTW: Are there (other) differences between Canadian French and European French that play a role for us?

cmfwyp:
I removed the *et al.* change.

@moewew 

Yes, `mathesis` should be translated to “mémoire de maîtrise” instead of “mémoire de master” in Canadian French.

Swiss French has at least one difference as well: the punctuation hooks for thin spaces and colon spaces should not be used.

Latin abbreviations are not related to European French or Canadian French. The *Lexique des règles typographiques en usage à l'Imprimerie nationale* and *Orthotypographie* recommend the French forms instead of the Latin abbreviations.

moewew:
Thank you. Maybe we should ignore the issue of Latin terms completely in *this* pull request and open up another one for that (that would also include `in`, `inseries`, `opcit`, `loccit`, `sequens`, `sequentes` besides the already discussed `andothers`). From time to time complaint about 'in'/'dans' comes up on TeX.SX, and often a native French speaker chimes in and says that using latinates such as 'in' is traditinal in French typography. (['Note in doesn't have to be translated: it' not the English word that's used, but the Latin preposition, and it is traditional in French bibliographies.'](http://tex.stackexchange.com/questions/303734/biblatex-in-text-in-bibliography-doesnt-get-translated#comment736400_303734)) So even while there seems to be advice to the contrary, using the latin terms seems to be traditional and/or still accepted in some fields.

It would of course have been convenient if the difference in Latin use was that easily explained (I noticed many of your links went out to `.ca`), but oh well...
I will see what I can do about a separate `canadien.lbx`. I don't see a way to split off a Swiss French file since `babel` has no special identifier for that variety.

cmfwyp:
That the Latin terms are traditional is true except for *cf.* (with the meaning *see also* rather than *compare to*) and *et al.* (the tradition in that case has been to enumerate all the authors without omitting any), but all sources I can find either don't recommend one over the other or recommend using the French forms, for sensible reasons. All the French forms are widely used today; arguably not as much as the Latin forms in scholarly publications, but the French forms can be understood immediately by any French speaker.

Another issue (avoided by using the French forms) is italics. There seems to be (unanimous) consensus (for French; I know *Chicago* says otherwise for English) that *op. cit.*, *ibid.*, *id.*, *loc. cit.*, and *sq.* (not *cf.*) must be put in italics. Is it possible or desirable to do this with `\mkbibemph`? As for "in", it is not put in italics in the examples I can find, probably because having the word "in" in italics just before a title in italics would be confusing. There's the same problem with *et al.* (which is one of the reasons Lacroux recommends against it) and *cf.* (which is probably the reason the *Lexique* does not put it in italics; I do not think *cf.* is more common than *ibid.*).

cmfwyp:
The latest commit has no changes related to the Latin abbreviations and I opened #555 (an issue rather than a pull request) to discuss the Latin abbreviations separately from this pull request.

moewew:
Thank you very much. I will merge this pull request shortly. If you want we can add your name to the list of language collaborators in the documentation.

perstar:
I think it's true for several languages that there is choice between using Latin or native terms for things like ibid, op. cit., etc. I know that when I helped with swedish.lbx it was sometimes hard to choose what Latin terms should be used. When I used Swedish terms for "idem" and "opcit" I added a comment about Latin also being good.

Maybe there should be a general option for encouraging/avoiding Latin terms, with latin=minimal/default/maximal or something like that. The settings for a language could give both a Latin and a native version for a term and state which should be the default unless you use the switch to explicitly say that you want as few Latin terms as possible (for example).

(Is this a worthwhile idea? Should I write a ticket?)

moewew:
@perstar It's an interesting idea, as I think in several languages there is choice between Latin or native words. Don't hesitate to add a ticket (if not for the sole reason that we have the discussion in a proper place and not attached to this pull request), but I'm not sure if we can follow up on it. Sometimes the use of Latin is stylistic, where the documentation already suggests to write a custom `.lbx` (pp. 244-245).

Issue #554, opened 2017-04-04
Dot missing after year in @periodical entry using authoryear style


ivalb:
Consider the following MWE:
```
\begin{filecontents*}{\jobname.bib}
@Periodical{jcg-2,
  Title                    = {BIS-Computers and Graphics},
  Subtitle                 = {A subtitle},
  Year                     = {2011},
  Number                   = {4},
  Volume                   = {35},
  ISSN                     = {0097-8493},
  Issuetitle               = {Semantic {3D} Media and Content}
}
\end{filecontents*}

\documentclass{article}

\usepackage[style=authoryear]{biblatex}
\addbibresource{biblatex-examples.bib}
\addbibresource{\jobname.bib}

\begin{document}

\cite{jcg} (from biblatex-examples.bib).

\cite{jcg-2} (from this file)

Missing dot after (2011) in the bibliography.

Subtitle missing for \textcite{jcg-2} in the bibliography.

\printbibliography

\end{document}
```

The dot after the year label is missing in both the entries inside the bibliography. And in the second entry the subtitle is missing too.

Ivan

moewew:
Mhhh this is more or less expected since we have `\setunit{\printdelim{nametitledelim}}\newblock`. So that delimiter can be overwritten by subsequent `\setunit`s. In this particular case it is overwritten by the first `\setunit*{\addspace}%` from `title+issuetitle`. I'm not sure if it is a good idea to use `\printunit` here.

The `subtitle` is not printed since the title was already printed and then cleared (to avoid repetition), but `periodical` reads
```
\newbibmacro*{periodical}{%
  \iffieldundef{title}
    {}
    {\printtext[title]{%
       \printfield[titlecase]{title}%
       \setunit{\subtitlepunct}%
       \printfield[titlecase]{subtitle}}}}
```

So far the analysis of the problem. I'm not sure yet if there is a good solution for any of these. I would welcome comments.

ivalb:
Considering that the problem occurs only for periodical entries and and only when the title is missing, a test could solve the issue. In fact, the title is not missing but it is cleared, so it is missing for the date+extrayear macro... The test prints a \newunitpunct and the subtitle (if defined).
```
\usepackage{xpatch}
\xapptobibmacro{date+extrayear}{%
\ifboolexpr
   {
    test     {\ifentrytype{periodical}}
    and test {\iffieldundef{title}} 
   }
   {\newunitpunct%
    \iffieldundef{subtitle}
    {}
[...]

PR #557, opened 2017-04-07
printunit instead of setunit for nametitledelim

moewew:
Cf. #554. Use `\printunit` instead of `\setunit` for `nametitledelim` as we probably always want it to prevail. 

Needs thorough testing as it affects many cases.

### DIFF
--- a/tex/latex/biblatex/bbx/standard.bbx
+++ b/tex/latex/biblatex/bbx/standard.bbx
@@ -27,7 +27,7 @@
   \usebibmacro{bibindex}%
   \usebibmacro{begentry}%
   \usebibmacro{author/translator+others}%
-  \setunit{\printdelim{nametitledelim}}\newblock
+  \printunit{\printdelim{nametitledelim}}\newblock
   \usebibmacro{title}%
   \newunit
   \printlist{language}%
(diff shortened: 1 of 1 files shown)


moewew:
I did a tiny bit of playing around with this and found that with really sparsely populated entries this can introduce different issues.

In a set-up with `backref` turned on and
```
\renewcommand*{\labelnamepunct}{\addcolon\space}
\renewcommand*{\newunitpunct}{\addcomma\space}
\renewcommand*{\bibpagerefpunct}{\addsemicolon\space}
```
an entry with only `title` and `year` will output
> A Book (2012): (cit. on p. 1).

where one could possibly expect (not too sure here)
> A Book (2012); (cit. on p. 1).

The latter is what we get with `\setunit` (status quo in dev).

So it seems that it is really hard to keep all corner cases happy.

Could you already run your test suite?

plk:
Do you still want me to merge and run the test suite for this?

moewew:
I think I need to revisit this. At the moment I would not merge as is.

moewew:
If you want to, you can close the pull request (without merging) and we can track this in #554

plk:
Ok, we'll do that

Issue #572, opened 2017-04-27
Normalise translatorstrng and translator+otherstrg


ivalb:
The _translatorstrg_ and _translator+othersstrg_ macros from _biblatex.def_ do not use any format to wrap the string. Suppose that we want to redefine _editortype_, for example, via
`\DeclareFeldFormat{editortype}{\mkbibparens{#1}}`
Of course this redefinition does not affect the entries with only a translator (plus _usetranslation_ option), like _vizedom:related_ from biblatex-examples.bib.
Since the _editor_ and the _translator_ has a similar role, bibliographically speaking, I would expect the same format. Here is a MWE:

```
\documentclass{article}
\usepackage[utf8]{inputenc}
\usepackage[T1]{fontenc}

\usepackage[backend=biber,style=authoryear]{biblatex}
\addbibresource{biblatex-examples.bib}

\DeclareFieldFormat{editortype}{\mkbibparens{#1}}

\begin{document}

\textcite{vizedom:related,gaonkar}

\printbibliography

\end{document}
```

The possible solutions to this issue could be:
1) to define a new field format: _translatortype_
2) to use _editortype_ inside the _translatorstrg_ and _translator+othersstrg_ macros:
```
\renewbibmacro*{translatorstrg}{%
  \printtext[editortype]{%
    \ifboolexpr{
      test {\ifnumgreater{\value{translator}}{1}}
      or
      test {\ifandothers{translator}}
    }
    {\bibstring{translators}}
    {\bibstring{translator}}}}
\renewbibmacro*{translator+othersstrg}{%
  \ifboolexpr{
    test {\ifnumgreater{\value{translator}}{1}}
    or
    test {\ifandothers{translator}}
  }
  {\def\abx@tempa{translators}}
  {\def\abx@tempa{translator}}%
  \ifnamesequal{translator}{commentator}
  {\appto\abx@tempa{co}%
    \clearname{commentator}}
  {\ifnamesequal{translator}{annotator}
    {\appto\abx@tempa{an}%
      \clearname{annotator}}
    {}}%
  \ifnamesequal{translator}{introduction}
  {\appto\abx@tempa{in}%
    \clearname{introduction}}
  {\ifnamesequal{translator}{foreword}
    {\appto\abx@tempa{fo}%
      \clearname{foreword}}
    {\ifnamesequal{translator}{afterword}
      {\appto\abx@tempa{af}%
        \clearname{afterword}}
      {}}}%
[...]

PR #573, opened 2017-04-27
new translatortype wrapper (#572)

moewew:
Cf. #572.
This introduces a format called `translatortype` analogous to `editortype` for `translator...` macros.


plk:
I'm alright with this but does it need adding to the docs too? If so, can you add it and add to the release information at the bottom of the file for 3.8?

moewew:
I can't even find the `editortype` format being defined in the docs. Where would we put it?

plk:
Oh, there is probably no need - it's likely covered by the statement that formats with the same name as the fields are automatically used.

moewew:
Yes, that makes good sense.

Issue #373, opened 2016-03-09
Spurious comma with authortitle style and \textcite


dbitouze:
Since `biblatex` 3.4, `authortitle` style and `\textcite` leads a spurious comma between the author(s) name(s) and the title. For instance, the following MWE:

```
\documentclass{article}
\usepackage[style=authortitle]{biblatex}%
%
\addbibresource{biblatex-examples.bib}
%
\begin{document}
\textcite{knuth:ct:a}
\end{document}
```

leads to:

> Knuth, (_TEXbook_)

whereas, as shown [here](https://www.overleaf.com/read/zyrjtppzjssq), it used to lead to:

> Knuth (_TEXbook_)

(until to `biblatex` 3.3).

moewew:
This is because in https://github.com/plk/biblatex/commit/8bf55f292832b2907435a80a8a29c0bd857b4c04 following https://github.com/plk/biblatex/issues/363 the `\textcite` commands were changed to use `\nametitledelim\bibopenparen` instead of `\addspace\bibopenparen`, this can cause problems if `\nametitledelim` contains anything other than a space. We already have `\extpostnotedelim` so that should probably be used here (and renamed?)

u-fischer:
@moewew I'm not sure if a new command is the solution. It can get confusing if one has a \nametitledelimbib and \nametitledelimtextcite and more variants.

"nametitledelim" is a clear name and if would a pity if one couldn't use it in all places where is makes sense. Imho one should ponder over a more generic solution. Perhaps one could define a test which allows to check if one is in a bib, or a textcite or a parencite so that one could do something like this:

~~~~~
\renewcommand\nametitledelim{\ifin{textcite}{\addspace}{\addspace\addcomma}}
~~~~~

Another idea would be a hook in the precode of a cite command so that one can add specific redefinitions.  

Or one could write something like this

~~~~~
\DeclarePunctuationFormat[textcite,parencite]{nametitledelim}{\addspace}
\DeclarePunctuationFormat[bib]{nametitledelim}{\addperiod}
~~~~~

and then use `\printpunct{nametitledelim}`. (But I'm not sure if it is a good idea to extend the Formatsystem from entrytypes to "citetypes". Perhaps it will confuse everyone ...).

moewew:
[...]

PR #575, opened 2017-04-28
Make more *namedelims context sensitive

moewew:
Make more delimiters use the context-sensitive `\printdelim`. 
Cf. #572, #373.

The commands changed are ubiquitous, so this needs testing. First tentative playing around with this revealed no immediately obvious problem.


moewew:
Played around with this a bit more and it still seems fine. Have not yet managed to run the test suite.

moewew:
Played around with this a bit more and it still seems fine. Unfortunately, I still can't run the test suite. Did your tests reveal any problems?

plk:
No problems with the test suite

moewew:
Brilliant. I hope to run a few more test over the weekend. If things go alright and you agree, I will merge then.

moewew:
I did some manual testing with randomly selected custom styles from CTAN and could observe no difference in output.

So if you don't see any reason against it, I think we can merge this.

PR #579, opened 2017-05-03
add \DeclareDelimAlias

moewew:
Add a way to alias delimiters to other delimiters declared with `\DeclareDelimFormat`.
This copies over all existing contexts.

Use as `\DeclareDelimAlias{finalnamedelim}{multinamedelim}` to make all instances of the `finalnamedelim` behave like `multinamedelim`.


plk:
Doc says both "delim" and "delimiter"?

moewew:
So it does. Sorry. Changed both mentions to `delim`.

PR #583, opened 2017-05-14
Update Brazilian translations

moewew:
See discussion in https://github.com/plk/biblatex/commit/b72238bcc449c32cac3043a397511987dae66a56

Will be mereged after @gusbrs has reviewed the changes.

### DIFF
--- a/tex/latex/biblatex/lbx/brazilian.lbx
+++ b/tex/latex/biblatex/lbx/brazilian.lbx
@@ -97,14 +97,14 @@
   founders         = {{fundadores}{fund\adddot}},
   continuator      = {{continuador}{cont\adddot}},
   continuators     = {{continuadores}{cont\adddot}},
-  collaborator     = {{colaborador}{col\adddot}},
-  collaborators    = {{colaboradores}{cols\adddot}},
+  collaborator     = {{colaborador}{colab\adddot}},
+  collaborators    = {{colaboradores}{colab\adddot}},
   translator       = {{tradutor}{trad\adddot}},
-  translators      = {{tradutores}{trads\adddot}},
+  translators      = {{tradutores}{trad\adddot}},
   commentator      = {{comentador}{coment\adddot}},
-  commentators     = {{comentadores}{coments\adddot}},
+  commentators     = {{comentadores}{coment\adddot}},
   annotator        = {{anotador}{anot\adddot}},
-  annotators       = {{anotadores}{anots\adddot}},
+  annotators       = {{anotadores}{anot\adddot}},
   commentary       = {{coment\'ario}{coment\adddot}},
   annotations      = {{notas}{notas}},
   introduction     = {{introdu\c{c}\~ao}{introd\adddot}},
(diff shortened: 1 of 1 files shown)


gusbrs:
@moewew , in re-examining `portuguese.lbx`, according to our conversation, I had a second thought regarding one single string in `brazilian.lbx`. Sorry about that. I had some doubt about the string `issue`, I had found a good solution for it, but `portuguese.lbx` inspired an even better one. If it is still possible (it is not critical at all), I suggest the change of the following string in `brazilian.lbx` to:

    issue            = {{n\'umero}{n\'um\adddot}},
 
Thank you.

moewew:
@gusbrs Note that `issue` is not `number`. `number` is the go-to subdivision of a `volume` of a journal (even if some journals would call that 'issue'). `issue` is normally only used for special issues of a journal or for non-numeric subdivisions: such as 'Winter issue'.

The docs say:
> The issue of a journal. This field is intended for journals whose individual issues
are identified by a designation such as ‘Spring’ or ‘Summer’ rather than the month
or a number. Since the placement of issue is similar to month and number , this
field may also be useful with double issues and other special cases.

gusbrs:
@moewew , yes I understand it, and that's precisely the nuance I think works better with this change. There are two words in Portuguese that translate from "issue" in English (in this context, of course). "edição", which was the original value of `brazilian.lbx`, which would be more related to a book and not properly to a journal. And "número", which works best for a journal. But "número" means also "number". The "number" (1,2,3) short form is usually `n.\textordmasculine`. And it is precisely the difference you mention which makes me think that "núm." is a better short form for "número" (meaning "issue", and not "number"). Still, how would `biblatex` expand the combination of the entryfield `issue` with the bibstring `issue`? Would it write simply "primavera número"? (If this is the case, it wouldn't make much sense in Portuguese, which would require something like "número de primavera" (different word order, plus preposition)).

moewew:
@gusbrs OK. As far as I can see the `issue` bibstring is not used at all in the standard styles, so we should probably not worry too much. I will change the string as requested.

gusbrs:
@moewew , Ok. Thank you very much, once again.

gusbrs:
@moewew , as you certainly noticed, I've been "looking into things" this week, and one of the things I observed is that the abbreviated string for `number` which I suggested last year when I looked into the strings for Brazilian (`n\adddot\textordmasculine`) is typographically quite ugly. So I came back to this matter.

I did the choice then based on what I could find around in the Internet about it. I couldn't find then any authoritative source, but did the best I could with what I found. But I happen to have found now a better source. I just stumbled into a norm of the Brazilian Association for Technical Norms (the infamous ABNT) about abbreviations (NBR 10522). This particular norm is old, and the own site of ABNT says it is not in application (but doesn't say either that it has been substituted for a newer version). Anyway, it seems pretty much to correspond to the actual use of abbreviations in the NBR 6023, which is the one currently "in force" for elaboration of bibliographical references in academic work. So, if it is not too much of a fuss, I'd be glad to take a new look at the abbreviated strings in `brazilian.lbx` in light of these norms. If you think it is a good idea.

moewew:
I always think it is a good idea if someone who actually knows about the localisation files has a look at them.

gusbrs:
Ok, will be done then.

gusbrs:
@moewew , you find an updated `brazilian.lbx` here: https://gist.github.com/gusbrs/6243f8bf64bce75daf08fd052e6332d3

The changes I introduced are mostly related to an update of abbreviated forms according to the norms I mentioned. They do not sanction an "s" to indicate plural in abbreviations. The cases which I had previously left with those were removed.

I also introduced another set of changes which have to do with the roles "annotator" and "commentator" when expressed as functions. In Brazilian Portuguese as they are, they sound pretty strange. To say "anotador" for this purpose in Brazilian Portuguese would sound as strange as "introductor" or "foreworder" in English. So I changed these for nouns, the same as was already done with "introduction", "foreword" and "afterword", and it sounds much better. It looks like this was directly translated from English in the past, and no one dared touch them afterwards (me included, the last time I checked these strings).

I also provided the few missing strings. Leaving only "candthesis" open.

moewew:
Thank you again.

I trust the changes from 'function' to 'result' make sense grammatically as used in the standard styles?

gusbrs:
Indeed. As I mentioned, "annotator" and "commentator" are now receiving the same grammatical treatment as "introduction", "foreword" and "afterword" in the roles expressed as functions. So they should work the same way as these last three do. Anyway, given your question, I did some tests to be extra-sure. And in fact they are not only correct, but imho much better.

Thank you!

gusbrs:
@moewew , I noticed some new strings for `langlatvian` and `fromlatvian` came about. I provide them for `brazilian.lbx`:

```
  langlatvian      = {{let\~ao}{let\~ao}},
  fromlatvian      = {{do let\~ao}{do let\~ao}},
```

Btw, please feel free to ping me when new strings are need for Brazilian.

moewew:
Done! https://github.com/plk/biblatex/commit/9e518687598a2f2613b0daa378589ef31f9036d0 Thank you very much.

Issue #520, opened 2016-12-21
Datetime. Missing frills and datetime output, authoryear style, and mergedate. Bugs.


JohnLukeBentley:
Some of the datetime frills (approximate, uncertain, eras) break in the authoryear style when mergedate is other than the default ("compact" or "true").

With ....

    alldates=ymd, 
    labeldate=year,
    mergedate=compact, 

... then output is as expected (bibliographic output in document, not alphabetic, order):

> * Alberici, Emma (2016). "Why Did 61 Million Americans Vote for Trump?" In: ABC News
> * ...
> * Kennett, Jeanette (2008). "True and Proper Selves: Velleman on Love". In: Ethics 118.2, pp. 213-227. issn: 0014-1704. doi: 10.1086/523747.
> * OED Online (2016). OED Online. Oxford: Oxford University Press.
> * ...
> * SEP (2016). The Stanford Encyclopedia of Philosophy. Ed. by Edward N. Zalta
> * Aquinas, Thomas (ca. 1273?). Compendium of Theology.
> * Plato (ca. 0380 BCE). Republic. Trans. by C. D. C. Reeve. 3rd edition. Indianapolis: Hackett
Publishing Company, Inc. 392 pp. isbn: 0-87220-737-4.

With 

    alldates=ymd, 
    labeldate=year,
    mergedate=basic, 

... then output breaks:

> * Alberici, Emma (2016). "Why Did 61 Million Americans Vote for Trump?" In: ABC News (2016-
11-20 21:02:28 +11:00).
> * ...
> * Kennett, Jeanette (2008). "True and Proper Selves: Velleman on Love". In: Ethics 118.2 (2008-01), pp. 213-227. issn: 0014-1704. doi: 10.1086/523747.
> * OED Online (2016). OED Online. Oxford: Oxford University Press, 2016-12.
> ...
> * SEP (2016). The Stanford Encyclopedia of Philosophy. Ed. by Edward N. Zalta.
> * Aquinas, Thomas (1273). Compendium of Theology.
> * Plato (0380). Republic. Trans. by C. D. C. Reeve. 3rd edition. Indianapolis: Hackett Publishing
Company, Inc. 392 pp. isbn: 0-87220-737-4.

This is broken in three ways:

1. The datetime frills (approximate, uncertain, eras) are absent from the first date; and
[...]

Issue #148, opened 2013-07-30
Add datelabel = year option setting


aboruvka:
Author-year styles don't consider month or days in the labels. Considering that, could a `datelabel = year` option setting be added? It perhaps should become the default given that the standard citation styles print only `labelyear`+`extrayear`.

Issue #601, opened 2017-07-30
Sorting of braced (corporate) authors with alphabetic style and BibTeX


moewew:
Via https://tex.stackexchange.com/q/372093/35864

Consider the MWE
```
\documentclass{article}
\RequirePackage{filecontents}

\begin{filecontents}{\jobname.bib}
@online{a,
author = {{Albert}},
title = {Title},
}

@book{b,
author = {Bernd},
title = {Title},
}

\end{filecontents}

\usepackage[style=alphabetic, backend=bibtex]{biblatex}
\addbibresource{\jobname.bib}
\begin{document}
\nocite{*}
\printbibliography
\end{document}
```

The sorting I get is
> [Ber] Bernd. Title.
[Alb] Albert. Title.

`Bernd` sorts before `{Albert}`. So it seems that with `alphabetic` braces are not stripped before sorting. 
Indeed, in the `.bbl` we get `\field{labelalpha}{Ber}` and `\field{labelalpha}{{Alb}}` as well as `\field{sortinit}{B}` and `\field{sortinit}{{A}}`. So `Bernd` sorts under `B` and `{Albert}` under `{{A}}`.

With `style=authortitle` the sorting is as expected. Sorting with Biber is good as well.

plk:
Perhaps @josephwright could have a look since it's about bibtex.

josephwright:
OK, I'll take a look :)

moewew:
We can almost get away with adding `purify` in

https://github.com/plk/biblatex/blob/8d925e4e3128fe82ea714ffa901ba07079d52ba6/bibtex/bst/biblatex/biblatex.bst#L2628-L2629

to make things read

            { labelalpha purify 'sortkey.alpha := }
            { labelalpha normalize extraalpha #4 pad.number * purify 'sortkey.alpha := }

That would certainly get rid of the unnecessary braces for sorting. But it also removes the `alphaothers` in case that is non-alphanumeric which could be a annoying for
```latex
\documentclass{article}
\RequirePackage{filecontents}

\begin{filecontents}{\jobname.bib}
@online{a,
author = {{Albert}},
title = {Title},
year = {1990},
}
@online{a3,
author = {{Albert}},
title = {Title},
year = {1989},
}
@online{a1,
author = {{Albert}},
title = {Title},
year = {1991},
}
@online{a2,
author = {{Albert} and others},
title = {Title},
year = {1990},
}

@book{b,
author = {Bernd},
title = {Title},
}

[...]

Issue #598, opened 2017-07-06
Write operation prevents font kerning


eg9:
Consider the following example

    \begin{filecontents*}{\jobname.bib}
      @Book{John,
        author = {John},
        title = {A Title},
      }
    \end{filecontents*}

    \documentclass{memoir}

    \usepackage[style=authoryear]{biblatex}
    \usepackage{Baskervaldx}

    \addbibresource{\jobname.bib}

    \showoutput

    \begin{document}
    (J)
    \autocite{John}
    \end{document}

The relevant part from the log file reveals

    ....\T1/Baskervaldx-TLF/m/n/10 (
    ....\kern1.62
    ....\T1/Baskervaldx-TLF/m/n/10 J
    ....\T1/Baskervaldx-TLF/m/n/10 )

for the `(J)` part, but

    ....\T1/Baskervaldx-TLF/m/n/10 (
    ....\write1{\abx@aux@page{1}{\the \c@page }}
    ....\T1/Baskervaldx-TLF/m/n/10 J

for the `\autocite`.

The `\write` operation prevents the font defined kerning. It should be performed before the parenthesis, which wouldn't change the reference (unless a break point is explicitly set before the name, which seems unlikely). The Baskervaldx font makes this necessity apparent, but other fonts might need a kerning between `(` and some letter.

Reference: https://tex.stackexchange.com/questions/378565/avoiding-glyph-collision-with-baskervaldx-and-biblatex

moewew:
I temporarily disabled the write, but still we don't get the kern. So I assume the problem is slightly more involved.

The log is
```
....\T1/Baskervaldx-TLF/m/n/10 (
....\kern1.62
....\T1/Baskervaldx-TLF/m/n/10 J
....\T1/Baskervaldx-TLF/m/n/10 )
....\glue 3.0 plus 1.49847 minus 1.00099
....\T1/Baskervaldx-TLF/m/n/10 a
....\T1/Baskervaldx-TLF/m/n/10 g
....\kern-0.12001
....\T1/Baskervaldx-TLF/m/n/10 a
....\T1/Baskervaldx-TLF/m/n/10 i
....\T1/Baskervaldx-TLF/m/n/10 n
....\T1/Baskervaldx-TLF/m/n/10 s
....\T1/Baskervaldx-TLF/m/n/10 t
....\glue 3.0 plus 1.49998 minus 1.0
....\T1/Baskervaldx-TLF/m/n/10 (
[...]

PR #605, opened 2017-08-06
Refactor mergedate settings and authoryear citations

moewew:
Cf. #520 (see also #148).

Currently mergedate is still slightly broken and the format between citations and bibliography differs in year labels. This refactor brings bibliography and citations in line by using the same commands and fixes most of mergedate's fixable behaviour.

We simply use `\printlabeldateextra` in citations, this greatly reduces clutter.

### DIFF
--- a/tex/latex/biblatex/bbx/authoryear.bbx
+++ b/tex/latex/biblatex/bbx/authoryear.bbx
@@ -66,7 +66,7 @@
       {\printdate}}%
   \renewbibmacro*{issue+date}{%
     \ifboolexpr{not test {\iffieldundef{issue}}
-                or test {\iflabeldateisdate}}
+                or not test {\iflabeldateisdate}}
       {\printtext[parens]{%
          \printfield{issue}%
          \setunit*{\addspace}%
--- a/tex/latex/biblatex/bbx/authoryear.bbx
+++ b/tex/latex/biblatex/bbx/authoryear.bbx
@@ -19,6 +19,13 @@
 
 \providebibmacro*{date+extrayear}{}
 
+\newcommand*{\iflabeldateisdate}{%
+  \ifboolexpr{%
+    not test {\iffieldundef{labeldatesource}}
+    and 
+    (test {\iffieldequalstr{labeldatesource}{}} 
+     or test {\iffieldequalstr{labeldatesource}{year}})}}
+
 \def\bbx@opt@mergedate@true{\bbx@opt@mergedate@compact}
 
 % merge date/issue with date label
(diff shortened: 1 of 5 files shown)


moewew:
Note that this is quite a large change. In my tests there were some differences in output compared to v3.7, but all of these were desirable as far as I can tell (and as far as I am concerned). I assume the test suite will also report different output, so that needs careful inspection.

plk:
Just ran the test suite and am checking output now.

plk:
First thing to note is the changed bibtex output where there seems to be repeated labelyear (from 01-introduction.tex):

screen shot 2017-08-06 at 5 00 pm

moewew:
Ah, OK I will have to have a look at BibTeX. The tests are based on `labeldatesource` which is much safer than just comparing labelyear and year with Biber, but I didn't think about BibTeX.

moewew:
If you can list all the other problems that arise I will have a look at those as well.

plk:
The only other difference I see for the biber tests is that entries with month/day in the .bib always have this now in the bibliography - I'm not sure about this:

screen shot 2017-08-06 at 5 20 pm

moewew:
This is my interpretation of what mergedate should do now. Previously entries like this would only have the month in the bibliography with the default `mergedate=compact` (this is the documented behaviour of mergedate in `50-style-authoryear`), but since we allow for days now we might as well include them.

The BibTeX problem could be solved by adding `labeldatesource` to the `.bbl`. I am currently trying to modify the `.bst` to do so. But I am quite clueless with `.bst` files, so if you have an idea...

plk:
Ok, yes, I think you are right. In that case, the biber test files look good - great work, many thanks. @josephwright - you are the bibtex expert here - can you help? We really would like to fix this soon so we can release.

moewew:
My attempt is at https://github.com/moewew/biblatex/commit/745596b90325f15cbc18cbb0183ef461d6d743cd

This adds `labeldatesource` with value `year` if labelyear is requested and we have a year.

moewew:
I also had to modify `labelendyear`, see https://github.com/moewew/biblatex/commit/ffab5a5a3851a34602c59443e3a3cc952d0e7699

moewew:
My suggested changes to `biblatex.bst` are in #606 now. It seems to work alright, but I would really appreciate someone with experience with BST files having a look.

plk:
Tests look good, I have committed the test changes and uploaded a new 3.8. Seems we are close to being able to release?

moewew:
Yes, but there are a few open issues left.
- #601 would be nice to have fixed before a release.
- I didn't get any further with #594, so unless you have an idea that would have to be deferred.
- #598, #586, #559 I'm clueless about and they are only 'small things', but can be _really_ annoying when noticed.
- #581 could hopefully be solved easily by someone who knows about `beamer`.
There is also #544 with `beamer` interaction.
- Regarding the date issue #540 is still open, and maybe John finds some quirks in the solution to #520 just committed.
- Not sure how we left #568.
- For weird punctuation problems there is #514 (we probably need to change the spacefactor of the ellipsis here) and #499.
- And #470 just crept up again.
- Also #585 was more complicated than I thought - mainly conceptually.

Of course that is just the list of issues that caught my eye when scrolling down the tracker.
#587 and #518 can easily be deferred.

plk:
@moewew, @josephwright - can we clean up the last few things soon - we really should release I think - it's a bit overdue.

moewew:
I think I could live with some of these things being deferred to the next version.
I started working on #585 and #514 and have at least a first attempt working. In #585 I would appreciate your feedback and ideally also Ulrike's response. In #514 I probably need Joseph to have a look.

PR #609, opened 2017-08-08
Fix typo

moewew:


plk:
Ah, I just corrected that too ...

moewew:
Oh, sorry. I'm too impatient sometimes.

plk:
Well, that's something we have in common then and I quite understand ... I appreciate the help - it is very good having someone else to help maintain the package.

PR #611, opened 2017-08-09
Implement subentry option for alphabetic styles

moewew:
While I experimented with how the `alphabetic` styles fared with the changes in #470, I noticed that they don't currently have a `subentry` option. Since that option also seems to make sense for `alphabetic` styles, here is an implementation.

While looking into documenting this in the style examples I found that the examples for `numeric` and `alphabetic` have lots of explanations only in the comments and not in the PDF. Should we look into making the text visible in the PDF?


plk:
Yes, I think it's best if the PDFs are self-contained as much as possible. I'll run the test suite and alter any tests which change as a result of this.

plk:
Tests look fine - I assume that there is nothing in the test files which would show up this change and so there are currently no conflicts. I think perhaps we might need to add a new test file for sets so we can check for regressions.

moewew:
Yes, the option is not activated by default. And I have not touched the style examples yet. So there should be no change. I can try to have a look at the style examples later on. I wanted to ask you about it before I went ahead.

plk:
The main difference is that `labelnumber` is generated by biblatex itself and assigned to set parents. `labelalpha` is generated by biber and is always a copy of the first set member `labelalpha`. I wonder if we might consider generating a `labelalpha` specifically for the set parents instead to bring it in line with `labelnumber` but this is not trivial because `labelalpha` generation is very complicated and customisable - it would be more so for sets because it would need to take account of multiple entries. Perhaps some sort of generic set label would be best?

moewew:
Obviously it would be great if there was some kind of automatic way of getting a label for the set parent. But I'm not sure what Biber should do to determine how that label should look. I think it is best to let the user set a label for the set parent manually with `shorthand`.

A generic 'Set' would also be possible but could potentially confuse users why things are labelled 'Set' when there is no 'set' to be seen in the data anywhere. But maybe that is just an academic question since I'm not sure how many people actually use sets with `alphabetic`.

plk:
Yes - I suspect bibtex uses the first member label as the set label too - might be just worth documenting in the alpha styles that the set label can be changed with shorthand but then that would also need to be present in the .bbl - I need to make biber do this after checking if bibtex does it.

moewew:
BibTeX failed to pick up `shorthand` in a test, but Biber used it to produce `labelalpha` which goes to the `.bbl`. Biber's behaviour is good here. Maybe we can do something about BibTeX.

plk:
Hmm, yes, biber does use the shorthand to generate labelalpha if it exists but it's the shorthand of the first set member. I wonder if we should allow sets themselves to have a shorthand and to use this if found for `labelalpha` and then fall back on some generic label for sets otherwise? This would decouple the set from the first member.

moewew:
Mhhh in 
```
\documentclass{article}
\usepackage[utf8]{inputenc}
\usepackage[ngerman]{babel}
\usepackage{lmodern}
\usepackage[T1]{fontenc}
\usepackage[style=alphabetic-verb,backend=biber,subentry=true]{biblatex}

\usepackage{filecontents}
@set{alphatest,
  entryset  = {sigfridsson,worman,geer},
  shorthand = {SWG},
}
\end{filecontents}

\addbibresource{\jobname.bib}

\usepackage{hyperref}

\begin{document}
\cite{alphatest}

\cite{sigfridsson,geer}

\textcite{sigfridsson,geer}

\printbibliography
\end{document}
```
I get 'SGW' for the set. That is the `shorthand` of the `@set` and not of any of its items.

plk:
Ok, good, I hadn't looked at that codepath for a while - must have already done this.

moewew:
Here is a suggestion to also let BibTeX accept `shorthand`s for `@set`s
```
FUNCTION {makelabel:ctrl:alpha} {
  ctrl:labelalpha
    { label empty$
        { push:cite purify #1 #3 substring$
          "u" change.case$ "t" change.case$ }
        { label makelabel:alpha:year * }
      if$
    }
    { shorthand empty$
        { "" }
        { shorthand }
      if$
    }
  if$
}
```
instead of https://github.com/plk/biblatex/blob/c345cac6b7f49937aeebef832850b729ca57ed24/bibtex/bst/biblatex/biblatex.bst#L2274

plk:
I'm fine with this if testing shows it works - it should be a relatively simple change since it's just adding a field to the .bbl.

moewew:
OK, I have committed that suggestion https://github.com/plk/biblatex/commit/4743fd016f5cb0a99cbfb7348ffdba5312666144.

A test revealed a further problem with BibTeX and `style=alphabetic`
```
\documentclass{article}
\usepackage[utf8]{inputenc}
\usepackage[ngerman]{babel}
\usepackage{lmodern}
\usepackage[T1]{fontenc}
\usepackage[style=alphabetic-verb,backend=bibtex]{biblatex}

\usepackage{filecontents}
@set{alphatest,
  entryset  = {sigfridsson,worman,geer},
  shorthand = {SWG},
}
\end{filecontents}

\addbibresource{\jobname.bib}

\usepackage{hyperref}

\begin{document}
\cite{alphatest}

\cite{sigfridsson,geer}

\textcite{sigfridsson,geer}

\printbibliography
\end{document}
```
I would expect that only the set appears in the bibliography (as is the case with Biber), but we also get to see the other entrysets.

plk:
Hmm, it looks worse for me - the set has nothing in the brackets at all, just `[]`.

plk:
Hmm, the set citing code is different between biber and bibtex. I am not so sure that this has been correctly split - it needs looking at.

moewew:
Yes, for some reason BibTeX fails to write `\inset` to entries in a set.

plk:
@josephwright - perhaps you can help with this? Seems bibtex output doesn't write `\inset` for set members any more? There was some code for inset in `biblatex.sty` which seems to belong in `blx-bibtex.def` and so I moved it there since it wasn't being referenced in `biblatex.sty` at all but that's nothing to do with this issue.

plk:
This is ok - remember that for bibtex, there has to be `\entryset` in all the set members too. If you do this, it works fine:

```
\documentclass{article}
\usepackage[style=alphabetic-verb,backend=bibtex]{biblatex}

\usepackage{filecontents}
@set{stdmodel,
  entryset     = {glashow,weinberg,salam},
  annotation   = {A \texttt{set} with three members discussing the standard
                  model of particle physics. The \texttt{crossref} field
                  in the \texttt{@set} entry and the \texttt{entryset} field in
                  each set member entry is needed only when using BibTeX as the
                  backend},
  shorthand = {SWG}
}
@article{glashow,
  entryset     = {stdmodel},
  author       = {Glashow, Sheldon},
  title        = {Partial Symmetries of Weak Interactions},
  journaltitle = {Nucl.~Phys.},
  date         = 1961,
  volume       = 22,
  pages        = {579-588},
}
@article{weinberg,
  entryset     = {stdmodel},
  author       = {Weinberg, Steven},
  title        = {A Model of Leptons},
  journaltitle = {Phys.~Rev.~Lett.},
  date         = 1967,
  volume       = 19,
  pages        = {1264-1266},
}
@inproceedings{salam,
  entryset     = {stdmodel},
  author       = {Salam, Abdus},
  editor       = {Svartholm, Nils},
  title        = {Weak and Electromagnetic Interactions},
  date         = 1968,
  booktitle    = {Elementary particle theory},
  booksubtitle = {Relativistic groups and analyticity},
  booktitleaddon= {Proceedings of the {Eighth Nobel Symposium}},
  eventdate    = {1968-05-19/1968-05-25},
  venue        = {Aspen{\"a}sgarden, Lerum},
  publisher    = {Almquist \& Wiksell},
  location     = {Stockholm},
  pages        = {367-377},
}
\end{filecontents}

\addbibresource{\jobname.bib}

\begin{document}
\cite{stdmodel}

\cite{glashow,weinberg,salam}
\textcite{glashow,weinberg,salam}

\printbibliography
\end{document}
```

moewew:
Ahhhh, that's why the `.bst` checked for `entryset` ... I hadn't read the annotation carefully. Should the documentation mention this (now that we basically dropped most of the Biber vs. BibTeX stuff)?

plk:
It does mention it once (3.14) but it used to mention it in the static sets section but no longer does. I'll correct this I think as it should be more explicit.

PR #624, opened 2017-09-14
Completed translation for Norwegian (Nynorsk/Bokmål)

hansfn:


hansfn:
Argh, the last commit message should of course be "Translate missing strings for Norwegian Nynorsk". Last time I tried to update a commit message for a pushed commit it didn't work out very well so I just leave it as is.

plk:
No problem, thanks for the translations.

PR #630, opened 2017-10-10
 Split postpunct from postcode 

moewew:
This came up in https://tex.stackexchange.com/q/334839/35864 a detailed analysis of what happens can be found in my answer there.

`biblatex` would gobble the postpunct after a cite command with only undefined citations. This is essentially due to the fact that both the postpunct and the general postcode of a citation command live in the same macro that is executed after a citation is processed. This command can not be (and is not) called after undefined citations. So for postpunct to work it needed to be split apart from the general postcode macro. This is what this pull request does.

Since this PR modifies low-level commands testing is needed. My manual tests as well as the OP's tests revealed no problems so far. Potential issue with this is that by splitting the postpunct from the postcode macro we cement the order of execution of postcode first, then postpunct, that was only implicit before. But I don't think that should be a problem.

### DIFF
--- a/tex/latex/biblatex/biblatex.sty
+++ b/tex/latex/biblatex/biblatex.sty
@@ -9982,7 +9982,7 @@
   \def\blx@dlimcode{\multicitedelim}%
   \def\blx@postpunct@saved{#6}%
   \ifblank{#2}
-    {\def\blx@postcode\@empty}
+    {\let\blx@postcode\@empty}
     {\def\blx@postcode{\usebibmacro{postnote}}}%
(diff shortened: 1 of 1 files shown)


plk:
I'm happy with this - splitting like this at least makes things explicit which makes it easier to adapt or change in the future.

moewew:
Thank you.

I was a bit unhappy with the name `\blx@postpunct@saved`, but the obvious `\blx@postpunct` was already taken. If you have a suggestion for a better name, that would be appreciated.

Issue #627, opened 2017-09-22
Uppercased marks with the standard classes


johannesbottcher:
Right now, the marks are `\MakeUppercase`d for the standard classes, whereas there are the right macros for KOMA classes (and memoir).

```
\documentclass{article}
\usepackage{biblatex}
\usepackage[markcase=used,automark]{scrlayer-scrpage}
\addbibresource{biblatex-examples.bib}
\begin{document}
\section{WomBat}\clearpage
\nocite{*}
\printbibliography
\end{document}
```

In the MWE above, the marks should be `used`, but biblatex sets them uppercased. 

Is there any chance to branch of a few more lines of code, or should that be better patched by the user?

Thanks

moewew:
What would you suggest here? Do we need to detect `scrlayer-scrpage`'s settings and act accorfingly or is there an easier way out?

johannesbottcher:
I fear it would be detecting the package. But `noupper` might just be what i did overlook. 

But then again, this is *template business* :-(

moewew:
Maybe someone who knows KOMA could have a look at our KOMA defaults https://github.com/plk/biblatex/blob/d0e866e1ee99027bbad5d530f8bd5b14f593fb6e/tex/latex/biblatex/biblatex.def#L1803-L1871 and tell us if things could be done in a better way.

moewew:
I never really got the hang of the KOMA packages - aside from happily using their defaults and the occasional high-level option.

Should our headings behave as they do for KOMA classes once `scrlayer-scrpage` is loaded or is there more to it?

johannesbottcher:
Hard week, i would need to look at it closer tomorrow. 

```
\documentclass{article}
\usepackage{biblatex}
\usepackage{scrlayer-scrpage}
\addbibresource{biblatex-examples.bib}
\begin{document}
\section{WomBat}\clearpage
\nocite{*}
\printbibliography
\end{document}
```
This MWE gives a header for the Refs, whereas none for the WomBat section, because headers weren't yet *demanded*. 
Unnumbered sections/chapters do not apply any `\mark*` commands by default (KOMA or standard), which is what most users do not expect. But if a mark should be set or not is a matter of opinion. 

`scrlayer-scrpage` is a package from the KOMA bundle, but can be used with the standard classes, which makes this a bit more tricky.

moewew:
A slight tangent, but while we are at it.

[...]

PR #631, opened 2017-10-15
Fix and simplify headings

moewew:
See #627 

- Introduces a new `\abx@MakeMarkcase` to control the format of running heads. Defaults to `\MakeUppercase` with standard classes, `\MakeMarkcase` if available (KOMA classes/packages) and `\memUChead` for `memoir`. 
- Use `\@mkdouble`, `\@mkright` for running heads (stolen from KOMA)
- If possible uses new `\bibliography@heading` with KOMA classes.

### DIFF
--- a/tex/latex/biblatex/biblatex.def
+++ b/tex/latex/biblatex/biblatex.def
@@ -1739,8 +1739,9 @@
 
 \defbibheading{none}{}
 
-\def\abx@MakeMarkcase\MakeUppercase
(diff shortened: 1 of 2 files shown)


moewew:
A few questions remain

- Should `\abx@MakeMarkcase` be a user-facing macro? (Should it get a name without `@`?) Does it need to be documented?
- Are `\@mkdouble` and `\@mkright` defined in the right place (`blx-compat.def` seemed like a good idea back then, but I'm not so sure any more)? Should they be documented?

Issue #657, opened 2017-11-23
Edited polish.lbx


acuiko:
Hi,
please find the edited polish.lbx file. I've checked all unsure/missing fields and edited some other
[polish_edited.txt](https://github.com/plk/biblatex/files/1498763/polish_edited.txt)

moewew:
Thank you very much.

Please check the changes in #658. I will merge the changes once you give the all clear.

PR #658, opened 2017-11-23
Update Polish localisation

moewew:
See #657.


acuiko:
I think it looks pretty good for now

moewew:
Thank you very much. I have merged the changes. This will be in the next `biblatex` release (the one after 3.9), since we only just released 3.9 it might take a while until this reaches CTAN.

Just open a new issue if something needs to be changed. We can list you in the section of localisation contributors in the documentation if you want to. I would need your name in that case.

Issue #659, opened 2017-11-24
biblatex unusable on upLaTeX


aminophen:
The change introduced in #506 is causing an error on uplatex (Japanese LaTeX); it seems that current biblatex is completely unusable on uplatex. An example in

https://en.wikibooks.org/wiki/LaTeX/Bibliographies_with_biblatex_and_biber#A_simple_example

can reproduce the issue by executing

~~~~
$ uplatex doc.tex
$ biber --bblencoding=utf8 -u -U --output_safechars doc
$ uplatex doc.tex
~~~~

~~~~
! Bad character code (8217).
~~~~

The list \DeclarePrefChars{'’-} contains ’ (U+2019), but (u)pTeX complains when the character is out of range of ASCII code in <code>\sfcode`(char)=(number)</code>. A workaround for Japanese users is to add \DeclarePrefChars{'-} to the preamble, but the situation is uncomfortable enough...

moewew:
Thank you very much for reporting this. And thank you also for pointing out the problematic bit of code as well.

@plk I think this means that at least the `.sty` files should be ASCII-only. At the moment the only problematic character is the `’` introduced in https://github.com/plk/biblatex/commit/aeb48bffe4dbb6f7cc80ba23952fb88cc7d461d1. We probably need to revert that change. This has implications also for #366. I suggest we restrict the defaults to ASCII-only and mention in the docs that non-ASCII chars need to be set by users (and that they can only expect those to work with a completetly UTF-8-aware engine).

josephwright:
You can go for a conditional: we have that in `expl3` for example for similar reasons. Test for the pTeX engines and skip over the 'unsafe' tokens?

aminophen:
> Test for the pTeX engines and skip over the 'unsafe' tokens?

I think it's good. However I'm not sure pdfTeX is really ok. Of course it is not causing an error, but is it UTF-8-aware and working as expected?

moewew:
Of course a conditional would be a nice idea. But is it safe to assume that (u)pLaTeX is the only engine that can't deal with this? And how do we safely test for pLaTeX (and upLaTeX?)? With `\kanjiskip`?

josephwright:
@moewew Testing for pTeX is easy: use `\kanjiskip`. For upTeX alone gets a bit more 'fun': see `l3sys.dtx`. Probably the safest approach is to have any skipped content at the end of a file (so you can `\endinput` for pTeX), or use a separate file.

moewew:
[...]

PR #664, opened 2017-11-28
Fix https://github.com/plk/biblatex/issues/659

aminophen:
All engines default to \DeclarePrefChars{'-} first, and XeTeX/LuaTeX are handled at the end of biblatex.sty. This should be safe for all known engines, including pTeX/upTeX/pdfTeX.

### DIFF
--- a/tex/latex/biblatex/biblatex.sty
+++ b/tex/latex/biblatex/biblatex.sty
@@ -1413,7 +1413,9 @@
     {\blx@defprefchars}
     {\let\blx@prefchars\@empty
      \blx@defprefchars}}
-\DeclarePrefChars{'’-}
+% Default for all engines; should be ascii-character only
+% For XeTeX/LuaTeX, the default is overwritten at the end of this file
+\DeclarePrefChars{'-}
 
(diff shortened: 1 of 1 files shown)


moewew:
Thank you very much for the suggested fix.

I'd prefer to have the non-ASCII chars in a dedicated file and not in `biblatex.sty`. If you want to amend the pull request to do that I will merge. Otherwise I will have a look at this later today.

aminophen:
> I'd prefer to have the non-ASCII chars in a dedicated file and not in biblatex.sty.

That would be nice. Then, "a dedicated file" means something like biblatex-utf8.sty?

moewew:
In analogy to other files, I would have gone with `blx-utf8.def`.

aminophen:
> In analogy to other files

Actually I don't know so much about the structure, so I'd like to leave it to you. Thanks.

moewew:
OK, I will close this pull request without merging then and will submit one later with a separate unicode file.

moewew:
If you can spare some time, it would be appreciated if you could have a look at #665 and do some tests to make sure this does not introduce additional problems for (u)pLaTeX.

Issue #656, opened 2017-11-21
Datetimes. The draft ISO 8601-201x and EDTF.


JohnLukeBentley:
## Intro

Biblatex formerly (between v3.5 and v3.8, inclusive) supported an "Extended Date/Time Format (EDTF)" specification (to "level 1") located at http://www.loc.gov/standards/datetime/pre-submission.html. A new International Standards Organisation (ISO) draft specification has emerged that incorporates, with minor changes, that specification and "supersedes" it (thanks @moewew for flagging the change).

I've read the recent draft ISO and would like to clarify the differences between the old EDTF spec, former and current biblatex support, and the new draft ISO. 

@plk indulge me addressing you in the third person even though you'll be the chief audience member in virtue of being the person who has implemented all this datetime stuff ...

@plk, with laudable enthusiasm, has gone ahead and [started implementing the draft ISO for v3.9](https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/CHANGES.md). @plk's partial implementation of the draft ISO spec has established that backwards compatibility is impossible (at least, and probably at most, with respect to open and unknown date ranges). And it seems that @plk has broken backwards compatibility in other ways on the basis that ....
>  [conforming to the old EDTF spec] syntax [was] too messy to do so and few people were using this anyway since it is a relatively new and specialised feature. https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/CHANGES.md

I think that's reasonable. However, I'd would have suggested holding off for now, or at least, for now, having biblatex not break the old EDTF spec. Given that the new draft ISO warns "it is subject to change without notice" and that, indeed, it is we who might very well want to lobby ISO to make such changes.

Therefore, for biblatex, given the option between:
* Waiting until the draft ISO is ratified before making the jump from the old EDTF; and
* Chasing the changing draft ISO until ratification.

... I suppose we are going to roll with the latter, having started down that path. That need not be too onerous: the ISO revisions are probably going to be slow and few. And I also suppose that I'm happy that @plk has been prepared to break backwards compatibility in all sorts of ways. For that better allows us to evaluate the new draft ISO as a general matter freed, to a greater extent, from parochial worries about biblatex. 

Anyway I suggest the purposes of this thread be:

* To clarify the former and current relationship between biblatex and the EDTF specs.
* To see if the new draft ISO throws up biblatex specific problems for us. 
[...]

PR #665, opened 2017-11-28
Use Unicode chars only with Unicode engines (#659)

moewew:
See #659, #664. Big thanks to @aminophen

Introduces a dedicated Unicode file that is only read by the Unicode-aware engines.
`biblatex.sty` is ASCII-only again.
`\DeclarePrefChars{'’-}` did not work as intended with pdfLaTeX anyway and did not throw an error only because the list loop could not handle UTF-8 characters properly.

### DIFF
--- a/doc/latex/biblatex/biblatex.tex
+++ b/doc/latex/biblatex/biblatex.tex
@@ -9993,6 +9993,11 @@ This command configures the \cmd{ifnumerals} and \cmd{ifpages} tests from \secre
 \begin{ltxexample}
 \DeclareRangeChars{~,;-+/}
 \end{ltxexample}
+
+On engines that fully support Unicode the default is
+\begin{ltxexample}
+\DeclareRangeChars{~,;-–—+/}
+\end{ltxexample}
 %
 This means that strings like <3--5>, <35+>, <8/9> and so on will be considered as a range by \cmd{ifnumerals} and \cmd{ifpages}. Non-range characters in such strings are recognized as numbers. So strings like <3a--5a> and <35b+> are not deemed to be ranges by default. See also \secref{bib:use:pag, use:cav:pag} for further details.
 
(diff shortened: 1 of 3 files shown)


plk:
Since I will have to release biber 2.10 shortly to address some further issues (this makes me realise that the year between 2.7 and 2.8 was just too long ...), can you do the merge into master instead of dev so that we don't have to release the ISO8601 stuff?

moewew:
I can do that if all goes well. I would like to wait a bit for feedback before merging anyway.

If there is nothing really pressing I would suggest to hold off releasing 3.10/2.10 for another week or two to see if nothing else turns up in the meantime.

plk:
Agreed.

moewew:
@plk I had a look at the EDTF/ISO changes. https://github.com/moewew/biblatex/tree/revertedtfiso is based on the current `dev` branch and reverts f91a917, https://github.com/moewew/biblatex/tree/reapplyisoedtf is based on `revertedtfiso` and reapplies the changes. If we were to merge https://github.com/moewew/biblatex/tree/revertedtfiso into dev we could go on developing at dev and if time comes for the EDTF/ISO changes we simply merge  https://github.com/moewew/biblatex/tree/reapplyisoedtf.

plk:
I’ve done things like this before and the danger is if something comes up before the remerge, this” gets messy. Given that branch names are arbitrary and that the main benefit is simply in having more than one branch, I’m not worried about this. In some ways I’d like to release the ISO changes soon since we are anyway currently running an non-standard draft which I’d rather not gave people adopt ...

moewew:
OK, I still think we could pull this off. Especially if we continually merge dev/master changes into the `reapplyisoedtf` branch the chances are low that we will end up with a huge mess if we try to merge the branches in the end. It might mean small bits of effort along the way, but I'd be happy to take care of that.

The merits of the approach are not very ample if the ISO changes are going to be applied soon. At the moment, however, the situation with the two branches is a bit unsatisfying since we have to cherry-pick around commits and that does no good to the commit history and mergeability either.

plk:
Do you have any objection to just releasing 2.10/3.10 with the ISO 8601 changes in a week or so? It would put things back to normal and I don’t see much of a problem since EDTF was always just a draft and all we are doing is moving to a better supported draft. It is unlikely to impact many or even any users as far as I can see.

moewew:
No, that would also be fine by me. It might be a bit annoying if we have to tweak this a bit later when the draft is finalised, but I believe that the features likely to change are nice enough that this does not cause too much upheaval. I thought you might have had an explicit reason not to release the ISO changes now, maybe in light of https://github.com/plk/biblatex/issues/656.

plk:
Ok, then we'll just use DEV as usual now and release from that fairly soon. I'm fine with making the ISO change now and tweaking as necessary, the novella of #656 notwithstanding.

moewew:
Relevant changes in https://github.com/plk/biblatex/commit/9ed28491871a131a874b5ff15e6b66575996edd1

PR #673, opened 2017-12-24
Add driver and biblistfilter to \printbiblist

moewew:
The `driver` and `biblistfilter` options would come in handy in https://tex.stackexchange.com/a/407352/35864.

Ideally, `\printbiblist` would also support a `sorting` option again, but that was not as straightforward as the other two options here.

### DIFF
--- a/doc/latex/biblatex/CHANGES.md
+++ b/doc/latex/biblatex/CHANGES.md
@@ -1,3 +1,7 @@
+# RELEASE NOTES FOR VERSION ??
+- `\printbiblist` now supports `driver` and `biblistfilter` options
+  to change the defaults set by the biblistname.
+
 # RELEASE NOTES FOR VERSION 3.10
 - **INCOMPATIBLE CHANGE** The recent ISO8601:201x standard supersedes
   the draft EDTF (Extended Date Time Format) extensions. Biblatex therefore
--- a/doc/latex/biblatex/README
+++ b/doc/latex/biblatex/README
@@ -162,4 +162,4 @@ Refer to biblatex.pdf for a systematic reference manual and to the
 CHANGES
 
 A list of changes relevant to users of this package is included at
-the end of 'biblatex.pdf'. See also the release notes in 'CHANGES.org'.
+the end of 'biblatex.pdf'. See also the release notes in 'CHANGES.md'.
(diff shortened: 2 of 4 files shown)


plk:
I'm a bit confused - `\printbiblist` does support the `sorting` option ...?

moewew:
I might be mistaken and I haven't checked thoroughly, but I think at the moment the sorting template for `\printbiblist` is fixed by the mandatory argument. As such it can be changed, but depends on the value of the mandatory argument, I would have found it nice if that could be overridden by an explicit `sorting` in the optional argument.

But this PR is only about the easy bits `driver` and `biblistfilter`.

plk:
It's not fixed - it defaults to the current refcontext and a mandatory argument but can be overriden by an explicit `sorting` option. Do you want me to merge this and add it to 3.10? It's no on CTAN yet because I'm still waiting for the last biber builder to upload so there is time.

moewew:
Mhh, to me

https://github.com/plk/biblatex/blob/e9c60eb44bb5f618feb7dd53273991a67be98859/tex/latex/biblatex/biblatex.sty#L8222-L8228

looked as though the sorting is fixed before the optional arguments are read. And `\define@key{blx@biblist2}{sorting}{}` made me think that `sorting` does nothing in printlists. But I'm not too sure about these kind of internals.

I'd like to merge the changes here, but whether they should make it to 3.10 or not I'm not too sure about. Since sourceforge has 3.10 as non-dev already, changing things would be cheating even if they are not on CTAN yet. So I'd leave that decision to you. The changes here are really not pressing at all.

plk:
I was not quite right - sorting is like with `printbibliography` - there is no option for this - it is taken from the surrounding refcontext but with `printbiblist`, the default is taken fro the mandatory argument and falls back on the refcontext value if there is no list matching the mandatory argument. Do you want a different behaviour?

moewew:
Mhhh, I thought it would be convenient if one could change the sorting as well. But from what I could see from the code that would have required more intricate trickery, so I didn't try to implement that. If you have a clever idea, go ahead, otherwise don't lose sleep over it.

Merry Christmas to you!

moewew:
Yes, the only way I could think of was essentially a third pass at reading the arguments before `biblist1` and `biblist2`, I was not too sure if that was worth it. If you don't have a clever trick around that it is probably not worth it.

Issue #675, opened 2017-12-30
Newer versions of biblatex fail to switch font encodings in a multilingual @entryset


alex-shpilkin:
After the recent CTAN update, the following no longer works ([test case](https://github.com/plk/biblatex/files/1594154/encofail.txt)):

- Import inputenc with `utf8`, fontenc with `T2A,T1` and babel with `russian,english`, setting the main document language to T1-encoded English with T2A-encoded Russian possible as an `otherlanguage`;
- Import biblatex with `backend=verbose-trad1,autolang=other,backend=biber`;
- Define two entries in the bibliography, one with `langid=english` and the other with `langid=russian` and Cyrillic characters, and an entryset containing both of them (in whichever order);
- Invoke `\parencite` inside the document to (try to) get an inline citation referring to the entryset.

The error message from pdflatex indicates that the font encoding was not switched from T1 to T2A. The encoding is switched correctly if an individual entry and not the entryset is cited.

moewew:
Thank you very much for reporting this bug.

For the benefit of the avid reader, here is the MWE again
```
\documentclass{article}
\usepackage[utf8]{inputenc}
\usepackage[T2A,T1]{fontenc}
\usepackage[russian,english]{babel}
\usepackage{csquotes}
\usepackage[style=verbose-trad1,autolang=other,backend=biber]{biblatex}
\usepackage{filecontents}

\bibliography{\jobname}
\begin{filecontents}{\jobname.bib}
@book{berezin1987introduction,
  title         = {Introduction to superanalysis},
  author        = {Berezin, F A},
  editor        = {Kirillov, Alexandre},
  translator    = {Niederle, Jiří and Kotecký, Roman},
  year          = {1987},
  origlanguage  = {russian},
  langid        = {english},
  publisher     = {Springer Netherlands},
  doi           = {10.1007/978-94-017-1963-6},
  isbn          = {978-90-277-1668-2},
}
@book{berezin2013vvedenie,
  title         = {Введение в~суперанализ},
  author        = {Березин, Феликс Александрович},
  note          = {Под~ред. Д.~А.~Лейтеса и А.\,А.~Кириллова с~дополнениями
                   Д.\,А.~Лейтеса, В.\,Н.~Шандера и~И.\,М.~Щепочкиной},
  edition       = {2},
  year          = {2013},
  langid        = {russian},
  publisher     = {МЦНМО},
  isbn          = {978-5-4439-0229-6},
}
@set{berezin-vvedenie,
[...]

PR #676, opened 2017-12-30
Add langbib to \blx@entryset (#675)

moewew:


moewew:
I'll be away for a few days so unfortunately I won't have time to improve this pull request in case it is not quite right.

PR #682, opened 2018-01-15
Make entrytype/entrykey available at bbl input

moewew:
Execute `\blx@bbl@data` just a tad later, so `entrykey` and `entrytype` are
already known.

Now the input handlers can filter by `entrytype` or `entrykey` with `\DeclareFieldInputHandler{title}{\ifentrytype{article}{\def\NewValue{}}{}}`. See https://tex.stackexchange.com/q/410384/35864

Preliminary tests have shown no problem with this change. But a run of the test suite would be appreciated.


plk:
Looks fine.

Issue #405, opened 2016-04-09
Biblatex 3.3: \entrydata* can't recover fields cleared by .bbx driver


dfussner:
When \blx@saveentry is called within \blx@imc@entrydata@i, usually at
the end of a .bbx driver, any fields cleared by that .bbx driver won't
receive "saved{field}" equivalents, making tests for them in the
related:* macros non-functional.  In related:multivolume, for example,
there's a test for whether "savededitor" and "editor" are the same,
but they can't be because "savededitor" is undefined, the "editor"
field of the main entry having been cleared in, I believe, all the
standard styles and probably most to all of the contributed ones.
This results in the "editor" being reprinted in all of the individual
volumes below the multivolume book, no matter whether it's the same or
a different editor in the relevant .bib entries.  My MWE and output
show this below.

If you move the \blx@saveentry call to the head of the driver, e.g.,
using AtEveryBibitem, then all of the fields are available for testing
in related macros, assuming those macros use the unstarred \entrydata.
Use of \entrydata* in the macro restores the current functionality,
with fields cleared by the .bbx driver no longer available.

Alternately, could one modify \clearfield in entries with a "related"
field, perhaps making it save cleared fields in another internal list
and having \blx@saveentry read that?

Here's the MWE:

\documentclass[12pt]{article}
\usepackage{csquotes}
\usepackage[american]{babel}
\usepackage[style=reading,backend=biber]{biblatex}
\usepackage{filecontents}
\begin{filecontents}{\jobname.bib}
@book{knuth:ct:a,
  author       = {Knuth, Donald E.},
  title        = {The {\TeX} book},
  date         = 1984,
  maintitle    = {Computers \& Typesetting},
  volume       = {A},
  editor = 	 {Report, Bug},
  publisher    = {Addison-Wesley},
  location     = {Reading, Mass.}
}

@book{knuth:ct:related,
  author       = {Knuth, Donald E.},
  title        = {Computers \& Typesetting},
  date         = {1984/1986},
  volumes      = 5,
  publisher    = {Addison-Wesley},
  location     = {Reading, Mass.},
  editor = 	 {Report, Bug},
  related      = {knuth:ct:a},
  relatedtype  = {multivolume}
}
\end{filecontents}
\addbibresource{\jobname.bib}
\begin{document}
\thispagestyle{empty}
\noindent
\nocite{knuth:ct:related}
\printbibliography
\end{document}

And here's the output on my system:

[...]

PR #684, opened 2018-01-18
Save date fields

moewew:
- Save date fields for `\entrydata*` (see https://tex.stackexchange.com/q/410977/).
- Correct `\blx@savebool`, the old macro saved nothing.
- Ping back to #405: `\abx@dointernalfields` now uses the normal
`\blx@savefield` and not the special `\blx@saverelfield`.


moewew:
@gusbrs Testing would be appreciated.

gusbrs:
@moewew , sorry but newbie question... I've tried putting the `biblatex.sty` in my tex file's folder and also copying the changed parts of the code to my file between `\makeatletter`\`\makeatother`, but I couldn't compile in either case. I suppose I'm missing other parts of the current development version which would be needed. Is that so? If so, what would be the proper way to have the development version for testing purposes only?

moewew:
@gusbrs Ah yes, you can't take the `biblatex.sty` from `dev` and drop it in your folder. Try https://gist.github.com/moewew/ddfe1f784f26aea924e5b69558d014dc.

The files provided on `dev` need to be 'installed' using the build script, see https://github.com/plk/biblatex#installation. For a quick test that might be undesirable.

gusbrs:
@moewew , I feared that much. But indeed currently I wouldn't like to fuss with my stable version. I'm already struggling with my little project, and it would not be a good idea to add one extra layer of complexity on top of that. That said, thanks for the drop-in replacement. I can compile with it, and will do some testing and report here soon.

moewew:
@gusbrs If you are on a Unixoid machine the install script allows you to install the dev version from github directly. If you install to `TEXMFHOME`/`~/texmf/` that should be fairly safe. You can uninstall that local version later on using the build script with its `uninstall` option.

But since the changes here are fairly minute and to only one file, a drop-in-replacement is much easier to use and even less dangerous.

gusbrs:
@moewew , yes I am on a Linux machine. But I'd prefer the less intrusive method at the moment, if the two are equivalent. However, if you think that, from a testing point of view, it would be somehow better to do so on a full dev version environment, I'll try to arrange the installation.

moewew:
They are not exactly equivalent. For the tests that are relevant here the drop-in-method suffices. Of course I would not mind if you tested the full dev version, but the more relevant question at the moment is whether the fix here works for you.

gusbrs:
moewew, there is no need for a "battle of politeness" really. I'm actually heavily indebted to you this week (not to mention long term standing debt), so I'll be more than glad to help within my reach.
Anyway, I'm up and running "dev version 3.11 savedbools branch" here.

I just started testing and stumbled in some issues of a kind I thought best to iterate on this. As far as I understand the code, you loop through the datetypes. So I would expect do find all the `\printsaved<datetype>date`s. Indeed they are defined, but behaved accordingly only for `date`. The result is:

print1

I think it is perhaps best if you took a look at this before I proceed. Unless you think otherwise.

The code I used to produce this was:

```\documentclass{article}

\usepackage[english]{babel}
\usepackage[style=authoryear,
%	datamodel=datestest,
	date=comp,
	block=par,%
	]{biblatex}

\usepackage{filecontents}

@book{mybook,
  author = {Author},
  title = {Book Title},
  date			= {2018-11-07},
}
@book{mybook2,
  author = {Author},
  title = {Book Title for dateextra},
  date = {2018},
}
@customa{myentry,
  author = {Buthor},
  title = {My entry's title},
  subtitle = {Below comes the testing},
  date = {2017-04-05},
  urldate 	= {2003-04-05},
  eventdate	= {1994-02-09},
  origdate	= {1985-11-03},
  mydate		= {1974-12-20},
}
\end{filecontents}

\addbibresource{\jobname.bib}

\DeclareBibliographyDriver{customa}{%
  \printnames{author}%
  \setunit{\printdelim{nametitledelim}}%
  \printfield{title}%
  \setunit{\subtitlepunct}%
  \printfield{subtitle}%
  \newunit\newblock
  \entrydata*{mybook}{%
		\printsaveddate
		\newblock
		\printtext{x}
		\printsavedurldate
		\newblock
		\printsavedeventdate
		\newblock
		\printsavedorigdate
%		\newblock
%		\printsavedmydate
		\newblock
		\printtext{\mkbibdateshort{savedyear}{savedmonth}{savedday}}
  }%
 \newblock
 \printdate
}

\begin{document}
\cite{mybook,mybook2,myentry}
\printbibliography
\end{document}
```

moewew:
Yes that was an annoying typo that did not cause problems with `date`, but did with all other date fields. Thank you for catching it. https://github.com/plk/biblatex/pull/684/commits/1d65cae07297164d1d9092ef790768f6e33ee2f1 should fix it.

gusbrs:
Great, thanks! Now, for me to proceed testing reflecting this fix, is it enough if I clone anew the savedbolls branch and reinstall? (Please be patient, I never got the hang of version control, though it is high on my todo list for quite some time...).

moewew:
Fetching and merging the changes (something similar to `fetch savedbool` and `merge origin/savedbool`) and then reinstalling should be enough. If something goes wrong, cloning from scratch can get you out of the mess.

If all goes wrong you can simply change the two lines from https://github.com/plk/biblatex/pull/684/commits/1d65cae07297164d1d9092ef790768f6e33ee2f1 manually.

gusbrs:
Up and running again. Thanks!

gusbrs:
moewe, still iterating. I can't seem to find `\printsavedlabeldate` and `\printsavedlabeldateextra`. Besides that, `\printsaveddateextra` misses the `extradate`. (I can provide code if required).

moewew:
Yes, labeldate is special. I have added code to save it as well.

`extradate` is a bit more tricky. All macros that use it hard-code `\printfield{extradate}`, so there is no simple way to print the correct `savedextradate` at the moment.

gusbrs:
It's ok. I'm just being systematic. These are not "requests".

moewew:
It definitely makes sense to print the proper `extradate` info as well. I'll see what can be done.

moewew:
@plk Any idea about `extradate`. If I understand it correctly, `extradate` is actually `labeldateextra`. Sometimes it can be useful to print that field not only via `\printlabeldateextra` but also in `\printdateextra` (of course only if `date` is `labeldatesource`). But of course the fact that it is hard-coded makes it hard to print `savedextradate`.

gusbrs:
@moewew , I've done a set of more systematic tests, as requested. My approach was that of an end user: put it to work and see what it does. Unfortunately, much of the code is pretty much beyond my depth, so that's what I can provide.

I prepared a test file (below) trying to use date information from the enclosing entry in an entrydata group in every scenario I could conceive (but in this context only). That is `\printsavedate` and variants. I can report the following issues:

- the `dateextra` part is missing in `\printsavedate`s as already reported.

- circa and uncertain information also go missing with `\printsavedate`. But there is a more general issue there too, even outside `\entrydata` (that is, the normal `\printdate`s) dates marked `~`, `?` and `%` receive always and only "circa".

- `\printsavedate`s do not carry format directive of their counterparts. There is a workaround for that, which is to use `\entrydata*{\thefield{savedentrykey}}{\printdate}`. But that's not ideal.

- use of nested `entrydatas` works, or almost. For english it works. For french, ngerman, and brazilian, the date receives a negative sign every other level.

- most important of all is that the `\printsaveddate` interact badly with `dateuncertain=true` and `datecirca=true`. For every date option which is not `comp` or `terse`, the use of `datecirca` or `dateuncertain` will render the file uncompilable.

The test file used is the following:

```
\documentclass{article}

\usepackage[%
    english,
%    ngerman,
%    french,
%    brazilian,
    ]{babel}

\usepackage[style=authoryear,
    datamodel=datestest,
    date=comp, % year, short, long, terse, comp, ymd, iso
    urldate=terse,
    eventdate=iso,
    origdate=comp,
    mydate=long,
    labeldate=year,
    alldates=comp,
    datezeros=true,
    timezeros=true,
%    dateabbrev=false,
%    datecirca=true,  % <-
%    dateuncertain=true,  % <-
%    dateera=secular,
%    dateeraauto=0,
%    time=12h, % 12h, 24h, 24hcomp
%    alldatesusetime=true,
%    labeldateparts=true,
    block=par,%
    ]{biblatex}

\usepackage{filecontents}

@book{mybook,
  author    = {Author},
  title     = {Book Title},
  date      = {2018-11-07},
}
@customa{myentry,
  author    = {Buthor},
  title     = {My entry's title},
  subtitle  = {Below comes the testing},
  date      = {2017-04-05T14:34:00~},
  urldate   = {2003-06-08T12:54:00?},
  eventdate = {1994-02-09T03:23:00/1994-04-10T07:02:00},
  origdate  = {1985-11-03T08:39:00%},
  mydate    = {-1564-10-12T20:47:00},
  xref      = {mybook},
}
@customa{myentry2,
  author    = {Buthor},
  title     = {My entry's title for triggering dateextra},
  date      = {2017-04-05},
}
\end{filecontents}

\addbibresource{\jobname.bib}

\DeclareBibliographyDriver{customa}{%
  \printnames{author}%
  \setunit{\printdelim{nametitledelim}}%
  \printfield{title}%
  \setunit{\subtitlepunct}%
  \printfield{subtitle}%
  \newunit\newblock
  \entrydata*{\thefield{xref}}{%
    \printtext{\mkbibbold{date:}}
    \newblock
    \printtext{d:}
    \printsaveddate
    \newblock 
    \printtext{de:}
    \printsaveddateextra
    \newblock
    \printtext{ld:}
    \printsavedlabeldate
    \newblock
    \printtext{lde:}
    \printsavedlabeldateextra
    \newblock
    \printtext{mks:}
    \printtext{\mkbibdateshort{savedyear}{savedmonth}{savedday}}
    \newblock
    \printtext{mkl:}
    \printtext{\mkbibdatelong{savedyear}{savedmonth}{savedday}}
    \newblock
    \printtext{ind:}
    \entrydata*{\thefield{savedentrykey}}{\printdate}
    \newblock 
    \printtext{tm:}
    \printsavedtime
    \newblock
    \printtext{\mkbibbold{url date:}}
    \newblock
    \printtext{d:}
    \printsavedurldate
    \newblock 
    \printtext{mks:}
    \printtext{\mkbibdateshort{savedurlyear}{savedurlmonth}{savedurlday}}
    \newblock
    \printtext{mkl:}
    \printtext{\mkbibdatelong{savedurlyear}{savedurlmonth}{savedurlday}}
    \newblock
    \printtext{ind:}
    \entrydata*{\thefield{savedentrykey}}{\printurldate}
    \newblock 
    \printtext{tm:}
    \printsavedurltime
    \newblock
    \printtext{\mkbibbold{event date:}}
    \newblock
    \printtext{d:}
    \printsavedeventdate
    \newblock 
    \printtext{mks:}
    \printtext{\mkbibdateshort{savedeventyear}{savedeventmonth}{savedeventday}}
    \newblock
    \printtext{mkl:}
    \printtext{\mkbibdatelong{savedeventyear}{savedeventmonth}{savedeventday}}
    \newblock
    \printtext{ind:}
    \entrydata*{\thefield{savedentrykey}}{\printeventdate}
    \newblock 
    \printtext{tm:}
    \printsavedeventtime
    \newblock
    \printtext{\mkbibbold{orig date:}}
    \newblock
    \printtext{d:}
    \printsavedorigdate
    \newblock 
    \printtext{mks:}
    \printtext{\mkbibdateshort{savedorigyear}{savedorigmonth}{savedorigday}}
    \newblock
    \printtext{mkl:}
    \printtext{\mkbibdatelong{savedorigyear}{savedorigmonth}{savedorigday}}
    \newblock
    \printtext{ind:}
    \entrydata*{\thefield{savedentrykey}}{\printorigdate}
    \newblock 
    \printtext{tm:}
    \printsavedorigtime
    \newblock
    \printtext{\mkbibbold{my date:}}
    \newblock
    \printtext{d:}
    \printsavedmydate
    \newblock 
    \printtext{mks:}
    \printtext{\mkbibdateshort{savedmyyear}{savedmymonth}{savedmyday}}
    \newblock
    \printtext{mkl:}
    \printtext{\mkbibdatelong{savedmyyear}{savedmymonth}{savedmyday}}
    \newblock
    \printtext{ind:}
    \entrydata*{\thefield{savedentrykey}}{\printmydate}
    \newblock 
    \printtext{tm:}
    \printsavedmytime
    \newblock 
    \printtext{\mkbibbold{nested entrydatas:}}
    \newblock
    \entrydata*{\thefield{savedentrykey}}{%
      \printdate
      \newblock
      \entrydata*{\thefield{savedentrykey}}{%
        \printdate
        \newblock
        \entrydata*{\thefield{savedentrykey}}{%
          \printdate
          \newblock
          \entrydata*{\thefield{savedentrykey}}{%
            \printdate
            \newblock
          }
        }
      }
    }
    \newblock 
    \printtext{\mkbibbold{other:}}
    \newblock
    \printtext{\thefield{savedentrykey}}
  }%
  \newblock
  \printtext{\mkbibbold{out of the entrydata group:}}
  \newblock
  \printtext{d:}
  \printdate
  \newblock
  \printtext{de:}
  \printdateextra
  \newblock 
  \printtext{evd:}
  \printeventdate
  \newblock
  \printtext{ord:}
  \printorigdate
  \newblock 
  \printtext{ud:}
  \printurldate
}

\DeclareFieldFormat{eventdate}{\mkbibemph{#1}}

\begin{document}
\cite{mybook,myentry,myentry2}
\printbibliography
\end{document}
```

gusbrs:
@moewew , I've edited a little, please see update.

moewew:
@plk While checking this I came across `dateunknown` and `dateunspecified`. It seems these bools/fields are only defined, but never used. Can they be removed?

moewew:
@gusbrs Thank you for your report. I have tried to address some of the issues you mentioned.

At the moment
- `extradate` does not work and
- the format as defined by `\DeclareFieldFormat` is not retained

Testing is appreciated.

If you find more issues, things are slightly easier to debug and fix if there is a short MWE for each problem and not a long one for all issues. I'm not quite sure if I missed something here.

gusbrs:
@moewew , ok, I will retest things here with the new commits. As to the formatting directives, I found a simple solution in the meantime. `\DeclareFieldAlias{savedurldate}{urldate}` does work, and the same can be done to other dates, I suppose.

moewew:
Indeed, `\DeclareFieldAlias` works if you don't have type-specific formatting. If you have type-specific formats these need to be copied one by one.

moewew:
@plk I have renamed the macros from https://github.com/plk/biblatex/pull/684/commits/8f35d8a19d22e62a1cb2d6b10a8e8011df92276b a bit so it is slightly clearer what they do.
If you want me to revert any of the changes before merging, just give me a shout.

Do you have any comments about `extradate` and `dateunknown`?

plk:
I have pushed a change to remove those booleans which I think are artefacts of earlier date code work.

plk:
Can you explain the `extradate` issue a bit more?

moewew:
@plk  Sorry, I was confused, `dateunspecified` is documented and makes sense, it is just not used in the standard styles. But I think `dateunknown` is indeed an artefact. 

The `extradate` issues is that from a conceptual point of view `extradate` is tied to `labeldate` (in a way it is `labledateextra`). But it makes sense to print the extradate info also for other `\print...dateextra` macros if `labeldate` and `...date` coincide (think `mergedate`). The problem is that `\printfield{extradate}` is hard-coded into the `\print...dateextra` macros and can't be adapted for `saved...date`s. The whole thing is a bit complicated, but I thought I might ask if you have a clever idea.

gusbrs:
@moewew , I retested the file here with the new commits, and as far as my testing goes indeed the problems with `dateuncertain` and `datecirca` are all gone. Thank you very much!

Except the issues you already mentioned, only one other problem remains. I thought it was a problem with nesting, but not quite. When there is an actually negative date printed within the `entrydata` group all the other dates are scrambled. This issue was language sensitive before the commits. It did not occur with english, but did with french, ngerman and brazilian. Now it occurs with all four.

The MWE below allows you to reproduce the issue:

```
\documentclass{article}

\usepackage[%
    english,
%    ngerman,
%    french,
%    brazilian,
    ]{babel}

\usepackage[style=authoryear,
    dateera=secular,
    block=par,%
    ]{biblatex}

\usepackage{filecontents}

@book{mybook,
  author    = {Author},
  title     = {Book Title},
  date      = {2018-11-07},
}
@customa{myentry,
  author    = {Buthor},
  title     = {My entry's title},
  subtitle  = {Below comes the testing},
  date      = {-2017-04-05T14:34:00},
  xref      = {mybook},
}
\end{filecontents}

\addbibresource{\jobname.bib}

\DeclareBibliographyDriver{customa}{%
  \printnames{author}%
  \setunit{\printdelim{nametitledelim}}%
  \printfield{title}%
  \setunit{\subtitlepunct}%
  \printfield{subtitle}%
  \newunit\newblock
  \entrydata*{\thefield{xref}}{%
    \newblock
    \printtext{\mkbibbold{correct date for `myentry':}}
    \newblock
    \printtext{d:}
    \printsaveddate
    \newblock
    \printtext{\mkbibbold{nested entrydatas (all dates have inverted sign):}}
    \newblock
    \entrydata*{\thefield{savedentrykey}}{%
      \printdate
      \newblock
      \entrydata*{\thefield{savedentrykey}}{%
        \printdate
        \newblock
        \entrydata*{\thefield{savedentrykey}}{%
          \printdate
          \newblock
          \entrydata*{\thefield{savedentrykey}}{%
            \printdate
            \newblock
          }
        }
      }
    }
  \printtext{\mkbibbold{even in the first level, date is wrong (but value is different from nested):}}
    \newblock
    \printtext{d:}
    \printdate
  }
  \newblock
  \printtext{\mkbibbold{correct date for `mybook':}}
  \newblock
  \printtext{d:}
  \entrydata*{\thefield{xref}}{\printdate}
}

\begin{document}
\cite{mybook,myentry}
\printbibliography
\end{document}
```

plk:
Hmm, the problem currently with `extradate` is that it isn't even generated by `biber` for anything other than `labledate` as this is what is used for citations. Is there a case for generating `extradate` for non-label dates? Basically `biber` determine `labeldate` from `\DeclareLabeldate` and then generates `extradate` only for labeldate. The same applies for the `extradatescope`.

moewew:
@plk Indeed. I don't think it would make sense to generate `extradate` for anything other than `labeldate`. Just the fact that `extradate` is always fixed in the date printing macros makes it hard to get `savedlabeldate` going. Thinking about this again, I don't think it makes sense to alter the status quo. I have a work-around in mind for the saved dates: It's not elegant, but it will work.

plk:
Ah, I see. The problem as I recall is that the placement of `extradate` has to be in the date printing routines because it needs to be next to the `extradatescope` part (by default the `labelyear`). Having it independent means that it is almost impossible to place it correctly with all the new ISO8601 date elements. It is theoretically possible I suppose to re-write all of this but I tried several ways when implementing the EDTF code and it is quite hard to do any other way which doesn't horribly complicate everything.

moewew:
@plk The `extradat` is definitely in the right place (except in situations where there is no right place). But gusbrs is onto something else in the last MWE

The same issue can be reproduced with
```
\documentclass{article}

\usepackage[%
    english,
    ]{babel}

\usepackage[style=authoryear,
    dateera=secular,
    block=par,%
    ]{biblatex}

\usepackage{filecontents}

@customa{myentry,
  date      = {-0333},
  origdate  = {2018},
}
\end{filecontents}

\addbibresource{\jobname.bib}


\DeclareBibliographyDriver{customa}{%
  \printdate
  \newblock
  \printorigdate
  \newblock
  \printdate
  \newblock
}

\begin{document}
\cite{myentry}
\printbibliography
\end{document}
```

The problems is that `\ifdateera` has two meanings. For one it the generic command in `\print...date` as defined in `\blx@metadateinfo`, but then it is also the type-specific command for `date` that can be used in other contexts as well. If several `\printdate`s are executed one after another the two meanings interfere resulting in unwanted effects.

moewew:
I see two solutions at the moment:

1. Rename the generic `\ifdateera` command to something that can't accidentally be overwritten by the date-type specific commands. We would have to rename all similar macros as well. This could potentially cause backwards compatibility issues, but I don't think the new macros are widely-used yet.
2. Group `\blx@metadateinfo`. This would require a change to all date macros that use it.

A search in my texmf tree showed that even though the change in 1 is backwards incompatible it should be OK for all styles published on CTAN.

plk:
I'm a bit confused - what would you expect with that MWE? I get:

```
334 BCE
2018
334 BCE
```

which seems correct?

moewew:
Yes, the problem occurs only with the code here in the branch. I'll try to find out where the problem is.

moewew:
Ah, sorry for the confusion; got it. At the moment the special date macros in `english.lbx` don't use `\blx@metadateinfo`, so the problem does not occur with `english` or `american`. I have added the call to 
`\blx@metadateinfo` here as well, so I could reproduce the problem with `english` as well.

The MWE below reproduces the issue with `dev` and 3.10 because it uses a language different from `english`.

```
\documentclass{article}

\usepackage[%
    british,
    ]{babel}

\usepackage[style=authoryear,
    dateera=secular,
    block=par,%
    ]{biblatex}

\usepackage{filecontents}

@customa{myentry,
  date      = {-0333},
  origdate  = {2018},
}
\end{filecontents}

\addbibresource{\jobname.bib}


\DeclareBibliographyDriver{customa}{%
  \printdate
  \newblock
  \printorigdate
  \newblock
  \printdate
  \newblock
  \printdate
}

\begin{document}
\cite{myentry}
\printbibliography
\end{document}
```

plk:
Hmm, then I suspect the right thing to do is as you mentioned and brace scope the calls to this macro better if you have an idea how to do this?

moewew:
In a way I like the idea of renaming one meaning of `\ifdate` more. But scoping would also be a possible solution. We would then just add a `\begingroup` at the beginning of `\mkdaterangefull` and friends and an `\endgroup` at the end. The problem with grouping is that we would have to document that grouping is necessary so nothing spills out. If we just rename `\ifdate` we won't have to worry about that.

moewew:
My preferred solution would be to rename `\ifdate...` `\blx@metadateinfo` to `\ifcurrentdate...`, but that means we must forbid `currentdate` for date fields to avoid clashes again.

See https://github.com/moewew/biblatex/compare/savedbools...moewew:resctrictcurrentdate

plk:
Hmm, looking at the commit, I'm not keen on this `current` thing - I find it confusing and would prefer to scope it so that `\ifdate...` just point to the correct in-scope date test. It was supposed to work like this - that's why I did it like this. I don't think I quite understand the problem yet - where is the conflicting `\ifdate...` call?

moewew:
The issue arises as follows.

In `\blx@blxinit` `biblatex` defines `if...datecirca` and friends for all date types, in particular `\ifdatecirca` is defined.
When `\print...date` is called `\blx@metadateinfo` assigns `\ifdatecirca` the meaning of `\if...datecirca`. So in a call to `\printurldate` `\ifdatecirca` is actually `\ifurldatecirca`. But this assignment is never reset, so after `\printurldate` `\ifdatecirca` is `\ifurldatecirca`. If you now call `\printdate` `\blx@metadateinfo`  assigns to `\ifdatecirca` the meaning of `\ifdatecirca`, but `\ifdatecirca` is `\ifurldatecirca`. So this gives the wrong test.

The assignments of `\blx@metadateinfo` must be scoped.

moewew:
The grouping solution is at https://github.com/moewew/biblatex/compare/savedbools...moewew:groupifdate

plk:
Ok, I get it now, haven't looked at this code for some time. I'm happy to merge this, it's a cleaner solution I think and style authors will have to look at this sort of code anyway.

moewew:
OK, let me ponder this for a bit longer and investigate the impact for custom styles. If all tests I run go well I will merge this evening.

moewew:
@plk I got errors when compiling `96-dates.tex`. Apparently Biber still writes `#1dateunknown` to the `.bbl`, so I was wrong to suggest to remove it in `biblatex`.

It should probably be reinstated and documented. `#1unknown` is used to differentiate
```
  origdate  = {/1034},     % open
  eventdate = {*/1034}     % unknown
```
The new ISO syntax would be `*/1034` instead of `../1034`. So it seems this requires changes in Biber as well.

From the `.bbl` it seems that `/` and `*/` now have switched meanings. `2000/` means open range and `2000/*` unknown.

plk:
I have pushed a reinstatement of this bool. I should pay more attention but I am very busy currently.

plk:
Are you saying that biber 3.10 isn't doing the right thing for ISO8601? I believe I changed it to correspond to the 8601 syntax.

moewew:
The LOC has taken down the ISO draft (https://www.loc.gov/standards/datetime/), so I can't link to the pdf, but in a version I found

§4.5 has
> * Unknown.  Start or end date unknown.  The character ‘*’ (asterisk) may be used for the start or 
end date   to indicate  “unknown”.
> * Open Start or  date open.  The  start  or  end date  may be left blank, either because there is none  or for any other reason.

The `..` notation is mentioned in §4.4, but not part of Level 1.

Of course this is all in draft state so heaven knows what comes round in the actual norm.

Have you pushed the changes to GitHub?

plk:
See the comments in CHANGES.md for biber 3.10 - these were the changes required for the latest 8601 draft that was available. Just pushed the `dateunknown` fix.

moewew:
Yup sorry for the confusion, the changes you made conform to ISO/DIS 8601-2:2016(e) dated 2016-10-26, but are contrary to what I found in a working draft of ISO 8601 dated 2016-02-16.

So I guess everything is fine.

Is there a reason why it is `abx@bool@end#1dateunknown` while all other instances of enddates use `#1enddate`?

moewew:
I'll try to document `unknown` dates and provide tests similar to `\ifdatecirca` later.

plk:
Hmm, that looks like a naming inconsistency - feel free to fix that ...

moewew:
Will do. But it requires a tiny change in Biber as well.

plk:
I think that biber is consistent in using `#1{end}dateunknown`. Just let me know if you need anything changed.

moewew:
Weird, the Biber code looks as though it would already give `#1enddate`... So a change might really only be needed in `biblatex`.

plk:
Right - I think that it only needs biblatex change.

PR #691, opened 2018-01-23
Italian i18n

fargiolas:
Translate a couple of missing strings in italian.lbx


moewew:
Thank you very much. Merged.

PR #708, opened 2018-02-10
Add translations for Latvian

rhssk:
Incomplete list with Latvian translations. A lot of the missing ones depend on context so I'll try to add them after a more thorough inspection.

Additional info as requested in [submission guidelines](https://sourceforge.net/p/biblatex/oldwiki/Adding_lbx_Files/#checklist-for-submiting-a-new-lbx-file):
* What kind of dash do you use in number ranges like "35-50"?
  * \textemdash
  * No spaces around the dash

* What kind of dash do you use in date ranges like "1st January-3rd February 2009"?
  * \textemdash
  * No spaces around the dash

* In lists and enumerations, do you add a comma in addition to connectors like "and"?
  * Typically no but there is an exception if the `and` is repeated more than once, for example,
`apples, oranges and cherries`, `apples, oranges, and cherries and pears` and `apples, oranges, and cherries, and pears and grapes` etc.

* How do you format short ordinals? Are ordinals gender/number sensitive?
gender/number insensitive: 1. 2. 3. 4. 5. ...

* What's the gender of the word "edition"?
masculine

* What's the gender of the word "series" (of a journal)?
feminine

* What's the long date format?
For 31.01.2009 the correct translation would be `2009. gada 31. janvāris`

* What's the short date format?
31.01.2009

* After which punctuation marks do you capitalize a word?
".!?"
The first word after ":" is usually written in lowercase unless it's a direct speech and starts with a quotation mark, for example,
`Lācis teica zaķim: "Āpsis atkal izēda manus medus krājumus".`

By looking at other .lbx files I wasn't quite sure how should space be handled in shortened words?
Is it okay to just write "A\adddot B\adddot" or should it be "A\adddotspaceB\adddot"?

Please let me know if additional info is needed or something needs corrections.


moewew:
Thank you very much for contributing the Latvian localisation.

I'll merge this soon, but there are a few things to work out first.

1. We prefer ASCII-encoded files, so if the UTF-8 characters can be encoded into ASCII+macros that would be preferred (I can do that for the commit you have submitted so far).
 2. How exactly is the date format built? I assume 'gada' is year? Is there declension involved for 'gada' (depending on the year number). What about declension of the month name?
3. Are there no  spaces after dots for the short date?
4. `A\adddot B\adddot` will put *no* space between "A." and "B.", `A\adddotspaceB\adddot` will put a space. What is correct will depend on your language. The languages I speak usually require a space after the dot.
5. Of course we would like to list you as a localisation contributor. Do you want to appear in that list and if so, can you give us your name for the list?

rhssk:
1. Should be fixed in the last commit
2. Correct, `gada` is year and the declension doesn't change on different year numbers in the current context. For months it depends. If the date represents a date in general (`January 31, 2009` and not
`on January 31, 2009`) then the month names stay as they currently are defined in translation file.
Another way of explaining is that if the date can be used as an answer to a question "What?" then translations are correct. If the question is "When?" then the months should be in different declension.

3. No spaces, yes
4. Fixed in the last commit
5. Rihards Skuja

moewew:
Thank you.

There are a few other things.

1. Re 'gada' you say "in the current context". Is it different with lets say year 0 or year 1?
2. What about `urldate` where the string is something like 'last accessed on 1. January 2017'. Would the month get a different declension here?
3. What is the 12h time format?
4. What is the 24h time format? What are the separators between hour and minute, and minute and second?
5. What happens if you only have a month-year date like "March 2010".
6. What's with season dates like 'Summer 2012'?

moewew:
If you want to check your localisation file you can put https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/03-localization-keys.tex in the same directory as your `latvian.lbx`, change the language to `latvian` and compile it.

rhssk:
1. A mistake on my part. When using `gada` to refer to a specific day (with month and date), the declension wouldn't change, only month declension is dependant on context. And the year number won't affect `gada` declension.

2. Yes, in that case the question would be "When?" and all months would get different declensions.

3. We don't really use the 12h format unless citing an English source.

4. hours:minutes:seconds with ":" as the separator

5. That would usually translate to `2010. gada marts` (2010. year March). `marts 2010` doesn't sound quite right but at least the declension will be correct for March and other months as well.

6. Same case as above. Would usually write if differently but the declension will be correct with current translations.

rhssk:
I'll try my best. If the month (in form that answers to "What?") ends with `s` then "When?" can be answered by substituting `s` with `ā`. If the month ends with `is` then it would be substituted with `ī`.

* janvāris -> janvārī
* februāris -> februārī
* marts -> martā
* aprīlis -> aprīlī
* maijs -> maijā
* jūnijs -> jūnijā
* jūlijs -> jūlijā
* augusts -> augustā
* septembris -> septembrī
* oktobris -> oktobrī
* novembris -> novembrī
* decembris -> decembrī

It wouldn't be pretty but if these rules can't be enforced programmatically, wouldn't it be possible to declare new month names (for example. january_in) and use those in some specific bibliography strings like `urldate`?

moewew:
> It wouldn't be pretty but if these rules can't be enforced programmatically, wouldn't it be possible to declare new month names (for example. january_in) and use those in some specific bibliography strings like `urldate`?

That was pretty much what I was planning on doing. But even that needs a bit of work, I think. The current date interface doesn't allow for these kind of format changes.

But let's try and approach this from the other side. Would it ever make sense not to have a date as the answer to the question "when?" in the bibliography?

rhssk:
No, if the date is in the form to answer "What?", using it as an answer to "When?" would sound awfully wrong. Though I could change the question from "When?" to "What?". For example,  `urldate` could instead be translated as `pēdējais piekļūšanas datums: 2009. gada 31. janvāris` (in English it would sound like `the last access date: 1. January 2017`). In this case both parts would answer to "What?".

moewew:
So the date in 'Sigfridsson and Ryde (1998)' the answer to what? Same goes for the date in "Anne Elk: On the Theory of Brontosauruses. 1. January 1972"?

We could defer the question by noting that the default format for `urldate` is the short date format 12.10.2016. Does it when vs. what make a difference then?

moewew:
I have merged what we had so far. I have also added a first attempt to satisfy the easy requirements for date an language extras.

Is Latvian by any chance similar to Estonian or Russian when it comes to date declension?

rhssk:
For both of those examples either "when?" or "what?" would be acceptable since the name of author or paper doesn't give enough information to be certain about whether it should be followed by an answer to "what?" or "when?". If it contained something like "in 1998" then it would be clear that answer to  "when?" should be used.

Yes, short date would be an acceptable solution as well, since text in numerical form doesn't have to answer to a specific question.

moewew:
So would it be very weird to go for the 'when?' form for all dates?


Is Latvian by any chance similar to Estonian or Russian when it comes to date declension? Then we could just adapt the code from there.

rhssk:
"So would it be very weird to go for the 'when?' form for all dates?"
For the long format yes. Some would be correct, some wouldn't. The short form should be correct for any case though.

moewew:
1. I think the Russian localisation uses something like "access date:" to allow for the what? instead of when? form for URL date. So maybe Latvian should do the same to avoid the whole when?/what? malarkey.
2. In what cases would Latvian require when?-dates for our purposes?
3. What if a long date is year-only do we then go for '2010', '2010.' or '2010. gada'?
4. Can you explain again about dates with only month or only season? I would need to know about the long and short format and about both when and what form.

moewew:
5. Should we use non-breaking or thin spaces in the long date format or anywhere else? What about leading zeros?

rhssk:
Looks like Russian uses shortened month names for `date` and numerical date form for `urldate` [(https://i.imgur.com/ir3IEhc.png)](https://i.imgur.com/ir3IEhc.png). The numerical form would work for Latvian as well but shortened month names wouldn't since there are some months that can't be shortened.

1. Correct. That's sort of what I suggested some 10 comments above this one.

2. I went over `03-localization-keys.tex` compiled in German and the only keys that use month name seem to be `date`, `urldate` and `eventdate`. And they're only an issue when the date also contains day and month: `{2006-10-01}`. In that case the question should be "When?".
I noticed that `origyear` should probably get special treatment as well since in German `origyear = 1959` gets translated to "Zuerst publ. in 1959" which in Latvian would be "Pirmoreiz publicēts 1959. gadā".

3. Both `2010` and `2010. gads` would work.

4. Not quite sure what more I can add about months but for seasons it would go like this:
what? -> when?
vasara -> vasarā
rudens -> rudenī
ziema -> ziemā
pavasaris -> pavasarī


5. I don't recall there being any rules against breaking dates which contain words. For me personally it would make sense not to break `24. marts` into 2 lines but I can't argument that with an actual rule, so probably leave just regular, breaking spaces in long date format. I think that for the rest of translations regular, breaking spaces can be used as well.
No leading zeros anywhere as far as I know.

rhssk:
I did some additional corrections (`03-localization-keys.tex` is really useful) and currently I see only the following problems:
* [https://i.imgur.com/ovdIUbA.png](https://i.imgur.com/ovdIUbA.png)
There should be quotation marks in the place of question marks. The correct style for Latvian is `,,lorus ipsem''` ([https://i.imgur.com/AOvwbDb.png](https://i.imgur.com/AOvwbDb.png))

* [https://i.imgur.com/u8G69UP.png](https://i.imgur.com/u8G69UP.png)
There's too much white-space after 1960. Also it would be ideal if only year number (with no dot) is shown if only year is specified: `year = 1960,` If day and month are specified as well (`date = {1960-09-16},`) then year should be followed by `gada`. So maybe something like
`dateyear = {{}{}}`
and
`dateyear_long = {{gada}{gada}}`

* `urlseen` should have a dot after year [https://i.imgur.com/QvRz1E9.png](https://i.imgur.com/QvRz1E9.png)

* [https://i.imgur.com/iJSSDY6.png](https://i.imgur.com/iJSSDY6.png)
Shouldn't URL be replaced with the translation for `url`? I have the same result for different languages so maybe there's simply an issue with my environment.

moewew:
The date format I implemented has some problems with spaces, I have already fixed a few of those problems locally, I will look at the rest tomorrow (well ... today).

The quotation marks come from `csquotes`, see https://tex.stackexchange.com/q/166337/35864. You can submit the correct format to https://github.com/josephwright/csquotes/issues

moewew:
Please submit a new pull request for any additional changes you made. Since this PR was merged manually, they can't be added to this one any more.

rhssk:
Thanks, cscquotes was indeed the issue.

moewew:
Shortening month names to hide their declension can not be an option, since that can be turned off with `abbreviate=false`. It just so happens that the default is to shorten the strings.


1. I know, but back then I thought you were unhappy with that work-around, so I tried to explore different avenues to get the dates right. Since it has become apparent that it is not that easy to, I wanted to check back with you how bad that solution would sound in Latvian. You compared it to 'the last access date: 1. January 2017' in English, which I find quite awkward (not a native speaker), 'date of last access' would sound better, 'access date' would also be OK, I guess. I wanted to know how awkward that translation feels in Latvian.

2. All dates can potentially have months and days. It is only the role in which the date is used in a style that dictates which form it could take.

3. I lean towards using only '2010' since the '. gads' seems like unnecessary text to carry around, especially in author year citations where (Sigfridsson and Ryde, 1998. gads) just looks weird.

4. You offered "2010. gada marts" (2010. year March) and "marts 2010". I think you preferred the former, but I'm not sure, can you please confirm that? I assume this is the *what* form for long dates. The *when* form for long dates would then be '2010. gada martā' is that correct? What about the short form? '3/2010'?
Would season dates used as '2010. gada vasara' (*what*)/'2010. gada (vasarā)' (*when*)?

5. OK. So there can even be a line break between '2010.' and 'gada'? Should it at least be discouraged?
I found a Microsoft style guide for Latvian that suggests not to break anywhere in the date, that seemed to excessive for me. Does no leading zeros anywhere mean also no leading zeros for the short date format? (I gathered from your initial post that those do use leading zeros)?

---------

6. Can you explain how compact date ranges like '1.--6. January 2010' and '1. February--6. March 2010' would work for long and short dates? What about month only ranges like 'February -- March 2010'?

7. The white space should be fixed soon.

8. OK to the dot after year for `urldate`. Does that mean that all short dates should have a dot after the year? I thought the short date format was '31.01.2009' without a dot. Or is `urldate` special because of the full sentence like structure?

9. The standard definition for the URL field is `\DeclareFieldFormat{url}{\mkbibacro{URL}\addcolon\space\url{#1}}` So always uses an untranslated 'URL'. The `.lbx` contain a string for `url` that can be translated, though, but that string is not used by the standard styles. But the standard is untranslated 'URL'.

rhssk:
1. I just thought that the original idea with no ":" would look cleaner and sound better. But "pēdējoreiz aplūkots: 2009. gada 31. janvāris" would work as well. It's not perfect but it's much better than if different parts of the text answer to different questions.

3. Agreed

4. Yes, the former is preferred.
You're correct about the _what_ and _when_ forms.
I looked over some academic guidelines and it looks like despite my original comment there should be leading zeros in numerical dates. So in this case it would be "03.2010".
For seasons it's similar as with months. It depends on context. If the text is isolated (in parenthesis or the only text in sentence) then either _when_ or _what_ could be used. If it's part of a sentence then it depends on what question the rest of sentence answers.

5. I just don't recall hearing about any rules against it and couldn't find anything definitive online. Can't recall seeing any split dates in academic works though so I'm not 100% sure here. Discouraging it would be fine I suppose.
For leading zeros see 4. "6. January 2010" would look like "06.01.2010."

6. Ranges are usually written like "no 2010. gada 1. līdz 6. janvārim" and "no 2010. gada 1. februārim līdz 6. martam". But then again there would be issues with declension. I suppose something like "1.-6. janvāris, 2010" and "1. februāris-6. marts, 2010" would work as well and wouldn't have issues with declension.
For short dates it would go like "01.01.2010-06.01.2010" and "01.02.2010-06.03.2010"
For months it would be similar: "februāris-marts, 2010" and "02.2010-03.2010".

6. Ranges should be similar to how regular dates are written. For example,
"2010. gada 1.-6. janvāris" and "2010. gada 1. februāris-6. marts"

8. Short dates should have a dot after every number. What I meant was that if the year is alone, like in parenthesis or alone in sentence, then it shouldn't be followed by a dot. For example, `124106 (2006.)` should be `124106 (2006)`.

9. Thanks for the explanation

moewew:
1. No ":" is absolutely fine as well (I would prefer no ":" as well, normally the bibstrings don't carry around this sort of punctuation).

I'll try to implement the dates as far as I can get and we can discuss the results then.

moewew:
I have committed first bits of date stuff
```latex
\documentclass{article}
\usepackage[latvian]{babel}
\usepackage{csquotes}

\usepackage[style=authoryear, mergedate=false, urldate=terse]{biblatex}

\usepackage{filecontents}
@book{elk1,
  author         = {Anne Elk},
  title          = {On the Theory of Brontosauruses},
  date           = {1972-09-18},
  url            = {https://example.com},
  urldate        = {2015-09-18},
}
@book{elk2,
  author         = {Anne Elk},
  title          = {On the Theory of Brontosauruses},
  date           = {1973-09-18/1973-09-30},
  url            = {https://example.com},
  urldate        = {2016-09-18/2016-09-30},
}
@book{elk3,
  author         = {Anne Elk},
  title          = {On the Theory of Brontosauruses},
  date           = {1974-09-18/1974-10-29},
  url            = {https://example.com},
  urldate        = {2017-09-18/2017-10-29},
}
@book{elk4,
  author         = {Anne Elk},
  title          = {On the Theory of Brontosauruses},
  date           = {1974-09-18/1975-11-29},
  url            = {https://example.com},
  urldate        = {2016-09-18/2017-11-29},
}
@book{elk5,
  author         = {Anne Elk},
  title          = {On the Theory of Brontosauruses},
  date           = {1974-09},
  url            = {https://example.com},
  urldate        = {2016-09},
}
@book{elk6,
  author         = {Anne Elk},
  title          = {On the Theory of Brontosauruses},
  date           = {1974},
  url            = {https://example.com},
  urldate        = {2016},
}
\end{filecontents}
\addbibresource{\jobname.bib}
\nocite{*}
\begin{document}
\cite{elk1}
\printbibliography
\end{document}
```
now gives

moewew:
Please check the updated version. It is now possible to request *when* dates with `\aftergroup\lbx@lv@requestwhendate` in the bibstrings that precede the respective date.

```latex
\documentclass{article}
\usepackage[latvian]{babel}
\usepackage{csquotes}

\usepackage[style=authoryear, mergedate=false, urldate=comp, dateabbrev=false]{biblatex}

\usepackage{filecontents}
@book{elk1,
  author         = {Anne Elk},
  title          = {On the Theory of Brontosauruses},
  date           = {1972-09-18},
  url            = {https://example.com},
  urldate        = {2015-09-18},
}
@book{elk2,
  author         = {Anne Elk},
  title          = {On the Theory of Brontosauruses},
  date           = {1973-09-18/1973-09-30},
  url            = {https://example.com},
  urldate        = {2016-09-18/2016-09-30},
}
@book{elk3,
  author         = {Anne Elk},
  title          = {On the Theory of Brontosauruses},
  date           = {1974-09-18/1974-10-29},
  url            = {https://example.com},
  urldate        = {2017-09-18/2017-10-29},
}
@book{elk4,
  author         = {Anne Elk},
  title          = {On the Theory of Brontosauruses},
  date           = {1974-09-18/1975-11-29},
  url            = {https://example.com},
  urldate        = {2016-09-18/2017-11-29},
}
@book{elk5,
  author         = {Anne Elk},
  title          = {On the Theory of Brontosauruses},
  date           = {1974-09},
  url            = {https://example.com},
  urldate        = {2016-09},
}
@book{elk6,
  author         = {Anne Elk},
  title          = {On the Theory of Brontosauruses},
  date           = {1974},
  url            = {https://example.com},
  urldate        = {2016},
}
\end{filecontents}

\addbibresource{\jobname.bib}
\nocite{*}
\begin{document}
\cite[7]{elk1}
\printbibliography
\end{document}
```
now gives

aasljasdlkasd

moewew:
Added another method to force locative dates https://github.com/plk/biblatex/commit/b5f72c06f4b26638068d444e60fa18d9e869518b

Please have a look at the output of the two MWEs from above and tell me what you think. If you could have another look at #710 that would be appreciated as well.

rhssk:
Everything looks really good except the last entry (1974c) `urlseen`. Since the year number is together with another word, it looks weird if there's just a raw number. In that case the year number should be followed by a dot and "gadā", like this: "(aplūkots 2016. gadā)" which would answer to "when?".

The more I look at it, the more I keep thinking that the dates should not be broken. It looks especially weird when there's a date range that gets split and "--" is the last character on the line. Could it still be possible to add non-breaking space for dates and ranges? What I had in mind was, for example, "1974. gada 18. septembris--1975. gada 29. novembris" could only get split after "1974. gada" and after "1975. gada". Is that somewhat similar how it would be done in English?

I'll try to give another go to #710 later today.

moewew:
OK. We can force 'gadā' for long *when* dates (but not for what dates, right? and its 'gadā' with a macron, while all other dates have 'gada' without macron?).

But please note that the *when* forcing is not fully set up yet - only `urldate` is currently a *when* date. All other dates that need to be *when* dates would need to be set up for that.

I can have a look at line breaking. Too few break points can be problematic.
Regarding the em-dashes and breaking, I think we must bite the bullet and accept breaking at the dash. Consider a date such as
> 1973. gada 18. septembris—1974. gada 30. novembris
We should discourage a break between the year and 'gada' as they form one unit. We should also discourage a break between '18.' and 'septembris' since those two form another unit. So breaking seems acceptable between 'gada' and the day. If we now also don't allow breaks around the em-dash there are too few places for proper breaking.
This would be similar to what English and German do, they also allow breaks at the dash and keep day and month tied together.

rhssk:
Correct. It should be "gadā" only when there's only the year "skatīts 2010. gadā" and "gada" when there's information about month or day as well "skatīts 2010. gada septembrī" or "skatīts 2010. gada 1. janvārī". So the strings are "gad\={a}" and "gada" respectively.

Okay, if that's how it's done for other languages, then let's not reinvent the wheel and add non-breaking spaces just between "year_number. gada" and "day_number. month_name" as you said.

Issue #717, opened 2018-02-24
Adding Hungarian support


marczellm:
I have forked biblatex at https://github.com/marczellm/biblatex and started working on Hungarian support with the help of @priestoferis / [Bence Ferdinándy](https://github.com/priestoferis).

This issue is to track our progress.

moewew:
Great! Let us know if you need help or anything fancy.

The new language file checklist is at https://github.com/plk/biblatex/wiki/Checklist-for-submitting-a-new-localisation-file-(.lbx)
Dates can do much more than they used to, so we need to know about 'long' and 'short' date formats (for dates with year, month, day; as well as for those with no day or no month; and for dates with only a season component) as well as date ranges and compact date ranges.

You can have a look at https://github.com/plk/biblatex/pull/708 and https://github.com/plk/biblatex/pull/710 for an example how this went with Latvian support.

priestoferis:
Ok, I'm adding this here, we'll edit this as we go along.

## Bibliography Extras

Adding bibliography extras requires more in-depth knowledge of Biblatex's localization modules. It is usually best if you go over the checklist below and send the answers to the Biblatex developers along with the translations; i.e., you translate the `\DeclareBibliographyStrings` part but omit `\DeclareBibliographyExtras`. 

  1. What kind of dash do you use in number ranges like "35-50"? 
      * `\textendash`? `\textemdash`? hyphen? 
      * Are there any spaces around the dash? 
  2. What kind of dash do you use in date ranges like "1st January-3rd February 2009"? 
      * `\textendash`? `\textemdash`? hyphen? 
      * Are there any spaces around the dash? 
  3. In lists and enumerations, do you add a comma in addition to connectors like "and"? Examples: 
      * UK English: "apples, oranges** and **cherries": no comma, like UK English
  4. How do you format short ordinals? Are ordinals gender/number sensitive? Examples: 
      * German: 1. 2. 3. 4. 5. ... [= gender/number insensitive]: insensitive, like German
  5. What's the gender of the word "edition"? 
      * No genders in Hungarian
  6. What's the gender of the word "series" (of a journal)? 
      * No genders.
  7. What's the long date format?
     1. What happens if the day is missing from the date? 
     2. What happens if both day and month are missing?
     3. Does this day format use leading zeros?
     4. Please be specific about the spaces to be used (non-breakable, thin spaces).
[...]

PR #728, opened 2018-03-10
translated few untranslated finnish phrases

kristiank:
I'm still unsure about correct terminology (if there even exists such a thing), but Estonian phrases is far better than Finnish phrases.

### DIFF
--- a/tex/latex/biblatex/lbx/estonian.lbx
+++ b/tex/latex/biblatex/lbx/estonian.lbx
@@ -423,9 +423,9 @@
 % sectiontotals    = {{}{}},% FIXME: missing
 % paragraphtotal   = {{}{}},% FIXME: missing
 % paragraphtotals  = {{}{}},% FIXME: missing
-  in               = {{teoksessa}{teoksessa}},
-  inseries         = {{sarjassa}{sarjassa}},
-  ofseries         = {{sarjassa}{sarjassa}},
+  in               = {{teoses}{teoses}},
+  inseries         = {{sarjas}{sarjas}},
+  ofseries         = {{sarjast}{sarjast}},
   number           = {{arv}{arv}},
   chapter          = {{peat\"{u}kk}{peat\"{u}kk}},
 % bathesis         = {{}{}},% FIXME: missing
@@ -440,7 +440,7 @@
   version          = {{versioon}{versioon}},
   url              = {{aadress}{aadress}},
   urlfrom          = {{saadaval}{saadaval}},
-  urlseen          = {{k\"{u}lastas}{k\"{u}lastas}},
+  urlseen          = {{vaadatud}{vaadatud}},
   inpreparation    = {{ettevalmistamisel}{ettevalmistamisel}},
   submitted        = {{esitatud}{esitatud}},
   forthcoming      = {{tulevases}{tulevases}},


kristiank:
I will try translate more in due time.

moewew:
Thank you very much for your contribution. If you can find some time, could you comment on

https://github.com/plk/biblatex/blob/45a7dcd128589acdf76f89656ab881d2a00b0d33/tex/latex/biblatex/lbx/estonian.lbx#L77-L81

I don't actually think the language files should redefine the `in:` bibmacro in this way. Even if it is unusual for Estonian to put an "in:" before the journal in for articles.

kristiank:
The macro has been copied from the Finnish file. There are several different ways to format bibliographies in Estonian.
Where do you think these kind of things should be defined if not in the language files?

moewew:
I'd just leave it and let users sort these things out themselves. No other `.lbx` files apart from Finnish does this sort of thing and it leads to inconsistencies and could even overwrite user decisions. So if there is no grammatical/linguistic reason for it, I would really like to get rid of it.

kristiank:
As far as I understand, the macro leaves out the `in:` if the `entrytype` is `article`. Since the `in:` is defined as "teoses" (literally _in the work_). I am not sure if it is propriate or not to use this term for entries of type article.
I am curius how should users sort these things out themselves? I am very new to biblatex and latex in general.

moewew:
A user could use the method discussed in https://tex.stackexchange.com/q/10682/35864 for example.

I have to trust your judgement here. The original localisations of this package are focused on German, English and French. There it is possible, if slightly unusual, to use "in:" for articles as well. Now if "teoses" is an absolute no-go for `@article`s in Estonian and not just a bit unusual (as it is in almost all languages), we would have to think about that.

I am a bit sceptical of `.lbx` files redefining bibliography macros. That can have unintended or unusual consequences in certain edge cases. And in this case it makes things inconsistent between Estonian and Finnish and all the other languages. That's why I'm concerned with the `in:`.

sim642:
@moewew I strongly believe that "in:" hack should not be part of the language. Whilst the simple approach to get rid of "in:" altogether is to override the bibliography string, this hack in the estonian language makes that impossible. One has to resort to messing around with the bibliography macro instead. `@article` to me seems in no way special than other types, which usually don't need "in:" either. This hack just makes using the estonian language more difficult by requiring completely not obvious workarounds which are not required in any other language.

I agree that it should be left to the end user to do something like that themselves if really desired. Seems like someone has just submitted their personal stuff into the repository which doesn't necessarily suit everyone else.

moewew:
@sim642 Thank you for weighing in here.

My feeling is also that an `.lbx` files should rather not modify bibmacros. But as it turns out the localisation system and the way it is used in some macros make it hard to satisfy the needs of certain languages. The system was written from an English/German/French background and simply has no interface for some concepts of other languages. Particularly troubling are for example changes in the word order and use of declension. Especially the latter can mean that certain strings have no sensible translation as a single word, but should rather be realised by other means (say, by declining the following phrase). The localisation framework has no support for declension of field contents and also has troubles with empty string translations (they may appear to work, but can break in certain situations since parts of the code were written assuming that `\bibstring{<string>}` always has a non-empty output).

This means that suppressing the output of a string by redefining it to be empty is problematic and should be avoided. If it is to be avoided, the next best solution is to redefine the field formats and macros that use that string. But that has the counter-intuitive effect that these formats and macros can now only be modified by a user or style in an `\DefineBibliographyExtras`. We had to take a route like this for Hungarian, see https://github.com/plk/biblatex/issues/717.

As an aside: Another problem is that `.lbx` files are not just about translation. To a limited degree they also venture into the realms of style. Take `french.lbx` as an example, it changes the format of family names to small caps. Apparently that is such an integral part of French typographic traditions that it was felt that the `.lbx` should apply this definition. In German on the contrary, small caps are not entirely unusual or unheard of, but it would be a stylistic choice whether or not to use small caps for family names and imposing small caps on users by default would probably be seen as wrong.

For the record and for other people reading this. Assuming the aim is to get rid of "in" for all entry types, overriding the bibliography **string** `in` still works (but it is not something I would recommend at all since it can lead to problems with punctuation, try `\renewcommand*{\intitlepunct}{XX}\DefineBibliographyStrings{estonian}{in={}}`). What does not work immediately is modification of the bibliography **macro** `in:` (this is my preferred solution), it has to happen in an `\DefineBibliographyExtras{estonian}` (i.e. `\DefineBibliographyExtras{estonian}{\renewbibmacro*{in:}{}}`). This is not immediately obvious from the start and contrary to the behaviour of almost all other languages.

Anyway, I'm open for suggestions to improve the situation that do not require a rewrite of the entire localisation system.

josephwright:
@moewew All sounds reasonable to me (though I'm always a bit disappointed when style leaks into language settings: it's a long-standing issue and not `biblatex`'s fault).

Issue #726, opened 2018-03-06
number vs issue


moewew:
Cf. https://tex.stackexchange.com/q/418590/35864, https://github.com/retorquere/zotero-better-bibtex/issues/925, https://github.com/plk/biblatex-apa/issues/45

Currently the docs are quite strict about `number` being an integer. For the `number`s as they appear in `@article`s that seems to be a bit too restrictive. There are at least two cases where one would put something other than a plain integer there

1. Special issues identified with a letter and integer, e.g. 'S1' (https://github.com/plk/biblatex-apa/issues/45)
2. Number ranges such as '2-3' (https://github.com/retorquere/zotero-better-bibtex/issues/925)

At the moment the documentation seems to suggest to use `issue` for non-integer input, and this is what Zotero BBT does. I find this unsatisfying since the output in the standard styles is significantly different when using `issue` as compared to `number`. It has also been established that for most intents and purposes `number` is the correct field for the subdivision of a journal `volume`.

If `biblatex` and Biber were to accept non-integer values for `number` the two cases above would easily give the expected output. Cautious style developers could use `\ifnumerals`, `\ifnumeral` or `\ifinteger` if they want to make sure the output does not end up looking stupid if they do anything special to the number field.

The only downside to this that I can see is sorting. If `number` is treated as a string, sorting pure integer-valued `number` fields might not give the expected output. But no standard sorting schemes sort by `number`...

retorquere:
Given that people are already putting things like ranges in `number`, and biblatex already accepts non-numbers there, things wouldn't get worse by just making the docs reflect the implemented behavior, correct?

retorquere:
(although I'd then have no way to decide in what field to put the zotero data, but that's tangential to this issue)

moewew:
@plk What do you think about this? The only drawback to not requiring `number` to be an integer that I can see at the moment is sorting.

plk:
In fact, the entire sorting system was overhauled in a major way to use a more efficient and typed algorithm precisely to accommodate integer sorting for these fields. The field type for sorting can be modified by the user in the sorting template but I would like to have a numeric field. I can perhaps make some changes to accommodate ranges (sorting on the first number in the range) but for arbitrary non-numeric parts, I think a new field would be better. Numeric sorting is important and having a free-form field again reverts to alpha sorting which impacts performance and is much less algorithmically clean.

retorquere:
So does that mean the issue field is back in favor?

moewew:
[...]

PR #730, opened 2018-03-18
Turn number into a literal field (#726)

moewew:
This would officially turn number into a literal field.


moewew:
If you are OK with the proposed changes here, I will resolve the merge conflict and merge from the command line.

plk:
Ok for me.

PR #736, opened 2018-04-09
Update biblatex-examples.bib

moewew:
Currently there is only a fix for the spacing in 'TeXbook'. This will change the output in the test files.

There are a few other things we may want to look at: Many titles use title case, while other English titles are written in sentence case. Since the general advice is to always give titles in title case and use a style that converts to sentence case if needed, should `biblatex-examples.bib` follow that best practice. Or do we stick to the titles exactly as written on the work?

### DIFF
--- a/bibtex/bib/biblatex/biblatex-examples.bib
+++ b/bibtex/bib/biblatex/biblatex-examples.bib
@@ -684,7 +684,7 @@
 
 @book{knuth:ct:a,
   author       = {Knuth, Donald E.},
-  title        = {The {\TeX} book},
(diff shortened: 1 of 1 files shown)


moewew:
@plk Any comments? Should we merge this)? What about the sentence case vs Title Case issue?

moewew:
While I'm at it: should `knuth:ct` be a `@mvbook`? It seems to be the archetype of one.

plk:
These changes are fine with me.

moewew:
OK, I'll push a commit for the `@mvbook` and will merge this.

Issue #746, opened 2018-05-21
Error when using biblatex with slovene locale


8greg8:
If I use minimal document like this:

>  \documentclass[11pt,a4paper]{report}
> 
> \usepackage[utf8]{inputenc}
> \usepackage[slovene]{babel}
> \usepackage{biblatex}
> 
> \begin{document}
> 	test
> \end{document}

the document will not compile. Compilation stoped with 

> File: slovene.lbx 2018/03/04 v3.11 biblatex localization (PK/MW)
> ) 
> 
>  ! LaTeX Error: Missing \begin{document}

With trial and error I found out that the problematic line is `481` in `slovene.lbx`. If I comment it out, then everything works.

moewew:
Thank you very much for reporting this. The problem was actually a missing comma at the end of the line you identified as problematic. I have fixed this in https://github.com/plk/biblatex/commit/943d40c9804d39b386b06df26a0f58dd1c38ba40

Until a new version is released you can get the `.lbx` file from https://github.com/plk/biblatex/blob/dev/tex/latex/biblatex/lbx/slovene.lbx

moewew:
`biblatex` 3.12 has been released and is available in TeX live 2018 and MikTeX now.

The error in the Slovenian localisation module has been fixed, the example posted in the first message compiles fine now.

PR #747, opened 2018-05-21
Patch slovene.lbx

8greg8:
`slovene.lbx` file with line `481` not working. If commented out, it works


moewew:
Thank you. I fixed the problem in https://github.com/plk/biblatex/blob/dev/tex/latex/biblatex/lbx/slovene.lbx by adding the missing comma as explained in #746.

For future pull requests please note that development takes place on the `dev` branch, `master` is for the current release on CTAN. So pull request should be made against `dev` and not `master`.

Issue #669, opened 2017-12-06
Again problems with autonum


christianlupus:
I already found issue #383. According to that, biblatex+autonum should work together. Unfortunately this is not correct for my installation. I will give a very minimalistic example:

    \documentclass[a4paper,10pt]{article}
    \usepackage{biblatex}
    \usepackage{autonum}
    \begin{document}
    Test
    \end{document}

As soon as one comments out either of the packages, compiling is working well. For both packages I get a message in the logs:

	(/opt/hd/texlive/2017/texmf-dist/tex/latex/biblatex/biblatex.sty
	...
	(/opt/hd/texlive/2017/texmf-dist/tex/latex/biblatex/biblatex.cfg))
	(/opt/hd/texlive/2017/texmf-dist/tex/latex/autonum/autonum.sty
	(/opt/hd/texlive/2017/texmf-dist/tex/latex/etextools/etextools.sty
	(/opt/hd/texlive/2017/texmf-dist/tex/latex/etex-pkg/etex.sty

	Package etex Warning: Extended allocation already in use.
	(etex)                etex.sty code will not be used.
	(etex)                To force etex package to load, add
	(etex)                \RequirePackage{etex}
	(etex)                at the start of the document.
    
    ...
    
	! Package biblatex Error: Incompatible package 'etextools'.

	See the biblatex package documentation for explanation.
	Type  H <return>  for immediate help.
	...                                              
							
	l.5 \begin{document}

It seems that biblatex tries to load the etextools package at the end of the preamble at `\begin{document}`. Of course then the autonum package has already loaded etextools at this time.

I am using the actual TeXlive 2017 version of all packages (updated a few minutes ago).

moewew:
@plk Do you have an opinion about what we should do here?

`etextools` breaks some of `etoolbox`'s macros and thus our internals. egreg mentions on TeX.SX that `etextools` is buggy and abandoned by its author (https://tex.stackexchange.com/a/238553/35864, https://tex.stackexchange.com/a/285953/35864, https://tex.stackexchange.com/questions/220268/biblatex-and-autonum-dont-work-together/220277#comment517465_220268).

I'm not sure how much of a deal-breaker bailing out if `etextools` is used is for end users. So far we only had this issue and @dbitouze  #587 saying we are a bit rude doing this. Considering how generally people seem to suggest not using `etextools` that is probably not too many complaints to change my mind.

[...]

Issue #587, opened 2017-06-02
Act more robust with outdated package versions


moewew:
A common problem is that users have outdated versions of packages loaded by `biblatex`, this causes at times not immediately obvious errors.

A recent example is an outdated `xstring` package: https://tex.stackexchange.com/q/370927/35864

From experience the problem is very prominent with outdated `babel` (especially `\bbl@loaded`).

For the `\bbl@loaded` issue we could check if the command is defined before using it.

For `xstring` and other packages (not sure which) we could at least use the optional date argument to `\RequirePackage` so that a warning is produced, alternatively we can throw a real error  as is done for `etoolbox`.

plk:
I have added tests for `xstring` and `babel` - can you have a look? I chose the version of babel one after the deprecation notices for the package which should be ok. `xstring` hasn't changed for years as haven't the other packages I think.

moewew:
Looks good. Ideally we would have taken the version of `babel` that introduced `\bbl@loaded`, but I could not find out which version that was.

We probably need to add `etextools` to the list of incompatible packages. I don't quite understand what `etextools` does exactly, but it messes with list handling in such a way that some loops fail. This is a bit weird since I didn't think `etextools` would redefine `etoolbox` internals.

plk:
I couldn't find that version for babel either but everything before 3.9q is deprecated and I think it was before that. I tried to use `etextools` for some things but as you say, it does really strange things with lists. I added an incompat warning for it.

moewew:
It seems the problem with `etextools` is that it redefines `\forlistloop` from `\forlistloop{<handler>}{<listmacro>}` (`etoolbox`) to `\forlistloop{<listmacro or expanded list>}\do{<...#1...>}` (`etextools`). These two definitions are completely incompatible.

I have added a short line in the docs about this: https://github.com/plk/biblatex/commit/75f4f8232182e9bfe6cf729a44957feeb1e76738

moewew:
Here is an interesting one. 

```
\documentclass{article}
\usepackage{polyglossia}
    \setdefaultlanguage{english}
\usepackage[style = authoryear]{biblatex}

\addbibresource{biblatex-examples.bib}

\begin{document}
\cite{sigfridsson}
\end{document}
```
complains about
```
! Package biblatex Error: Outdated 'babel' package.
```
even though `babel` is not loaded. Does `polyglossia` pretend to be `babel`? Even if `polyglossia` should not do that, we probably need to drop 
[...]

PR #751, opened 2018-05-29
Check for package incompatibilities earlier

moewew:
If incompatible packages are loaded before `biblatex` they can cause problems already at the loading stage. To allow for more meaningful messages in some of these cases it makes sense to test for incompatible packages immediately at the beginning of the package.

A few definitions needed to be moved around.

**INCOMPATIBLE CHANGE** Because options are not set up that early, `noerroretextools` could not be honoured. This means that we need to go back to a more primitive approach of checking for the existence of `\blx@noerroretextools`.

This still does not catch some package incompatibilities if an incompatible package is loaded after `biblatex` and the mere loading of the two causes errors.

Ping back to https://github.com/plk/biblatex/issues/669 and https://github.com/plk/biblatex/issues/587

### DIFF
--- a/doc/latex/biblatex/CHANGES.md
+++ b/doc/latex/biblatex/CHANGES.md
@@ -5,6 +5,23 @@
   consider using a custom data model to turn `number` back into an integer type
   field, since sorting integers as literals has performance implications and
   might lead to undesired sorting such as "1", "10", "2".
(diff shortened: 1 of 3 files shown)


moewew:
@plk Can this be merged as is? This is going to be a breaking change for `noerroretextools`, but I think it is going to give the average user a better behaviour for incompatible packages to check at the start of the package.

plk:
Fine for me.

PR #763, opened 2018-06-16
Two fixes for BibTeX

moewew:
- Remove braces for sorting in alphabetic labels (see #601)
- Allow `shorthand` labels in alphabetic to have a `extraalpha` (consistent with Biber's behaviour)
- Zap tabs and spaces and end of line


josephwright:
Looks good to me.

moewew:
Thank you for checking. The test suite did not complain, so I'm merging this now.

PR #769, opened 2018-07-08
Catalan strings 2

sebasvila:
Some more fixes to the catalan strings. Added the japanese and ukrainian languages. 

While checked with 03-localization-keys,  maybe discovered a bug of \smartof. In reference 6 it says «Ed. i com. de Edward Meredith Cope.», however it would say «Ed. i com. d'Edward Meredith Cope.». I was unable to see the reason. It's better to post a bug?


moewew:
Thank you very much. Yes that looks like a bug in `\smartof`. I will open a bug report shortly.

moewew:
See https://github.com/plk/biblatex/issues/770

PR #775, opened 2018-07-24
Add fallback values for bibliography extras

moewew:
Provide sensible defaults (from `british.lbx` -- `english.lbx` was not viable due to the different American date format that would have required other ancillary macros as well ...) for date and time macros
that should be redefined by all `.lbx` files.

This seems slightly safer than swallowing dates silently. The definitions should not be needed
and the warnings should never be encountered with the 'official' `.lbx` files - but they may help to debug custom `.lbx` files.

Cf. https://tex.stackexchange.com/q/442587/


moewew:
Maybe the date should be in printed in bold to make the connection with missing bibstrings more obvious?

Issue #555, opened 2017-04-05
Use Latin forms with italics or French forms in french.lbx


cmfwyp:
See #552 for context.

moewew:
Regarding the use of italics for Latin terms normally I would prefer not to include `\mkbibemph` in the bibstrings themselves, but rather to redefine `\mkibid` which is used by all standard styles to typeset scholarly Latin abbreviations such as `ibidem` etc.

This would not work very well for 'et al.', though. Since there is precedence for putting `\mkbibemph` in bibstrings directly ([`biblatex-ieee` does this](https://github.com/josephwright/biblatex-ieee/blob/master/ieee.bbx#L59)) and since the usage of italics with Latin seems to be compulsory in French (and not just a style issue) I would probably be OK with putting `\mkbibemph` into the bibstrings directly.

I welcome comments on this issue as well.

moewew:
@cmfwyp @dbitouze  @Evpok Is there anything to discuss here? Or have you all found peace with the status quo?

Just to recap: The question is whether we should retain the Latin versions of certain terms or whether `french.lbx` should use the French versions of these. Examples being 'et al.' -> 'et collab.' A related question is about the technicalities of italics for Latin terms.

If these things are merely stylistic and both Latin and French terms are encountered in the wild, I would vote to retain the status quo.

dbitouze:
@moewew IMHO, a nice solution would be to implement the [@perstar suggestion](https://github.com/plk/biblatex/pull/552#issuecomment-291797109). Meanwhile, the status quo is OK for me. If you want, I could ask other opinions on French (LaTeX) forums.

moewew:
Mhhh, not sure if that can be implemented easily. The localisation system is quite involved. It seems to me that even though the Latin terms should be fairly universal, there are again subtle differences in usage (cf. #667). So ultimately the thing seems to be a stylistic question and I would like to avoid stylistic decisions if possible. The `.lbx`s are at an unfortunate intersection of translation and stylistic decisions. Implicitly they enforce stylistic choices, but we don't want to pick sides in general. In particular I don't want to start a 'war' between different stylistic factions. 

Thank you for the offer. I'm not sure if a large discussion is really needed, if people had serious problems with how things are at the moment they would probably have complained already. I'd like to let sleeping dogs lie.

cmfwyp:
[...]

PR #790, opened 2018-08-30
Estonian: added translations

kristiank:
Added some missing translations and fixed some old translations.

Please check whether \DeclareBibliographyStrings is well-formed, e.g. is the comma `,` allowed on line 582.

```
\DeclareBibliographyStrings{%
% ... snip ...
  am               = {{AM}{AM}},
  pm               = {{PM}{PM}},   % <---- is this comma supposed to be here?
}
```


moewew:
Thank you.

The comma is fine. It is optional after the last entry, but I strongly prefer to include it in case a new string is added at the end.

A compilation of the new file against https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/03-localization-keys.tex shows that

- `pagetotal`
- `pagetotals`
- `columntotal`
- `columntotals`
- `sectiontotal`
- `sectiontotals`
- `paragraphtotal`
- `paragraphtotals`
- `versetotal`
- `versetotals`
- `linetotal`
- `linetotals`
- `langjapanese`
- `langukrainian`
- `fromjapanese`
- `fromukrainian`

are missing. So if you could find the time to include those as well that would be great.

AlexDaniel:
Any news?

Also, I'm interested in a new release with this line fixed: https://github.com/plk/biblatex/commit/30d925d8b79cf08b96fdead66a31e2af0678bea3#diff-16af56c5dc46222e28d6ee7d5cceb4b6L407
Basically everything works fine on the `dev` branch but the fix is not in 3.11 release, so it fails for any user who attempts to use Estonian language.

kristiank:
@AlexDaniel what is wrong with line 407 in that file?
@moewew I'll add and correct more this evening.

AlexDaniel:
@kristiank currently nothing, because you fixed it in March. But the last [release](https://github.com/plk/biblatex/releases) happened before that fix.

kristiank:
Can I change the latin phrases into _italic_ and if so, how do I do that?

AlexDaniel:
@kristiank correct, it's @moewew or @plk or any other contributor with access who should cut a new release. Sorry for confusion that I caused in this PR, I should've started a new ticket for that.

moewew:
@kristiank Brilliant! Thank you. I'll wait with merging this pull request until then, so you can push changes to here.

Re the Latin phrases: Things are a bit complicated. See also https://github.com/plk/biblatex/issues/555. For the `verbose` styles there is a command `\mkibid` that can be used to format what is commonly a Latin phrase (`opcit`, `loccit`, `ibidem`, `idem`). But in general it cannot be guaranteed that the argument of `\mkibid` is a Latin phrase in your language translation (in the German translation `ibidem`, `idem`, `loccit` and `opcit` are all rendered in German and not Latin) and it is also not guaranteed that all Latin phrases are actually used within `\mkibid` (case in point: `andothers`, "et al.", is not used with `\mkibid`). So currently the only reliable way to have Latin phrases in italics is to add the respective command (`\mkbibemph` or `\mkbibitalic`) to the bibstring (as is done in `biblatex-ieee` for example: https://github.com/josephwright/biblatex-ieee/blob/f1607910b82373b51ecdd2d90a1f3ba5cfb1c875/ieee.bbx#L62). I'm not particularly fond of that because I believe that bibstrings should only contain translations and not additional markup and that ideally they should not enforce stylistic decisions. But if your language has strict rules that require the use of italics with Latin then this is the best option. If there are different opinions about whether or not Latin phrases need Italics, I'd prefer to come down on the side of not introducing additional markup in the `.lbx` file, but I'll defer to your judgment. That all said, there are more than enough example of `.lbx` files meddling with formatting (`french.lbx` the new `hungarian.lbx`) that my purist view proves to be too naive and outdated.

@AlexDaniel Yes, I guess enough things have built up that it makes sense to think about a new release. Unfortunately due to interdependencies with Biber pushing out a new release is not so simple and takes a while to prepare and pull off. PLK (who does the real work here) has been very busy in the last months, so I can't promise anything.

kristiank:
I'll wait with the latin formatting (the crux is, that in Estonian all foreign text should be styled in a marked way, e.g if the Estonian text is itself in italics then the foreign language text should not be in italics ...)

Can someone verify I didn't break anything with removing all occurrences of `\finalandcomma\` and other.

I found a lot of garble in the file and I can't translate it all in one go. Sorry.

moewew:
There are quite a few `\}`s that should only be `}`s (in general I guess it would be safe to assume you never want `\}` or `\{` in an `.lbx` file). This causes brace mismatches and errors along the way. It seems that the wrong `\` is only inserted after macros. Note that you only need ```\<space>``` (a control space) after a macro if you want to avoid that the following space is gobbled (cf. https://tex.stackexchange.com/q/31091/35864).

I could try and fix these syntax errors, but I'm not sure if I come across a situation where knowledge of Estonian is required to understand what a correct fix would look like, so if you could do it that would be appreciated.

You can check if your `.lbx` file works by compiling against https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/03-localization-keys.tex with your language enabled. Ideally there should be no errors there and all strings should look as intended.

Removing `finalandcomma` is fine if Estonian does not have the concept of an (optional) Oxford comma (see https://en.wikipedia.org/wiki/Serial_comma).

kristiank:
I fixed the backslashes by changing `\} -> }` and also removing backslashes before space. There's no optional Oxford comma in Estonian.

I wonder whether `\adddotspace` is valid? What difference is there to use `\adddot` instead of a plain `.`?

kristiank:
I now tried compiling `03-localization-keys.tex` and found the results very helpful. I will look it through and see if I can make the translations more correct. Since this will take time, I think it is ok to publish the estonian lbx as it is now (or including it in a release or however this is called) -- IF there is no errors of course.

moewew:
Thanks. Please note that removing `\<space>` after a macro changes the output.

Compare
```latex
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\newcommand*{\fakeaddot}{.}

\begin{document}
Hello\fakeaddot Goodbye.

Hello\fakeaddot\ Goodbye.
\end{document}
```

Usually LaTeX commands that don't take arguments gobble any following white space. See https://tex.stackexchange.com/questions/31091/space-after-latex-commands. That means that `\adddot` like the `\fakeaddot` in the example eats the space that comes after it. So when `\adddot` needs a space afterwards (this is not always necessary, for example when you write `\adddot,`) you need to 'protect' the space from being gobbled, this is done with `\<space>` ('control space').

So I think that https://github.com/plk/biblatex/pull/790/commits/e2c1dead50b9102edab0c803cd72fa67db0ce6b5 removed too many backslashes. Have a look at the short form of `translatorco` for example.

`\adddotspace` is a valid command, it is different from `\adddot` in that it automatically also adds a space, but that space has a slightly lower chance of allowing a line break than `\adddot\ `. The `.lbx` files are not really consistent in their use of `\addotspace` vs `\adddot\ `, so if you don't feel comfortable with one you can use the other.

Re the Latin stuff again: If you use `\mkbibemph` the emphasis toggles between italics and upright (if it is used in an upright context, the Latin form would become italic, if the surroundings are already in italics it would switch to upright).

I will wait to merge until you have had a look at the space changes. But please note that as I write to @AlexDaniel even if we merge the pull request here an actual (official and public) release of the package so that downstream users can use the new localisation will take some time and I can give no promise of an ETA at this time. Sorry.

kristiank:
Thanks for clarifying. I will add appropriate spaces this evening.

moewew:
@AlexDaniel I know, but unfortunately the situation is as it is.

moewew:
There should be a space in `U\adddotS`

```
U\adddot S
```
or even simpler
```
U.S
```
because in this special case `\adddot` has no real benefit over a literal `.` (this also holds for all situations where `\adddot` is directly followed by a comma).

You will also never want to have
```latex
\adddotspace,
```
as in `byeditortrin` and a few other strings. 
```
.,
```
would be enough there or
```
\adddot,
``` 
if you prefer that (again the first is simpler and is at no disadvantage compared to `\adddot,` in these situations).

There might also be a problem with `\lblx@lfromlang`/`\lbx@sfromlang`.
- If a work has been translated (and `origlanguage` is defined), then `\lbx@lfronlang` holds the string `from [the] <language>` so that you can use it to build `translated from the <language> by ...`.
- The macro `\lbx@sfromlang` should be used in the short form of the string, the long form should use `\lbx@lfromlang`. It seems that currently all strings use `\lbx@lfromlang`.
- I also noticed that the position of the macro with respect to `` is not consistent.
We have `t\~{o}lkinud \lbx@lfromlang` but `toimetanud, \lbx@lfromlang t\~{o}lkinud ja kommenteerinud` Unfortunately, the test file does not populate the macro.

Take the modified example file from https://gist.github.com/moewew/78da26715a5c2b76b033c7f2dfd21d69 that pretends that a work has been translated from the French. You will want to check all strings that output the word "prantsuse".

kristiank:
All should be done now. I made a digital printout of the `03-localization-keys.tex` and will go through it "sometime when I have borrowed an appropriate dictionary or textbook from the library". Thus I think this pull request be merged and wait for future work. Hopefully a new release will reach Estonian users in due short time.

Thank you very much for helping me a lot.

moewew:
Thank you very much. I have squashed the commits before the merge to keep the history shorter, so you only see one commit https://github.com/plk/biblatex/commit/625b9139f18ff4ce95b4cd42965177851051a7e8 that implements this.

PR #800, opened 2018-09-20
Estonian

kristiank:
I found help and wordlists with local terminology. Many terms clash (theoretically), but I guess that it is very uncommon that they clash in the local practice.

My understanding of the `fromlang` functions is that they add a space if necessary, so I haven't added any spacing myself.


kristiank:
Sorry. I still get all these old commits coming up here. I did try to `fetch` and `merge`. Also I created a new dedicated branch for my estonina translation editing.

moewew:
Yeah, I guess we still get that because I squashed your earlier commits. I think they are on your `dev` branch, but since I squashed them they are technically not on our `dev` branch (our `dev` branch only has the resulting squashed commit). Unfortunately I don't really have the git skills to recommend a way out for you that would get your `dev` back to our `dev`.

moewew:
It is a bit unfortunate that there are so many strings with the same translation now, but if you think that this is the best way to go and that the older translations were unnatural then I'm fine with that.

There are tree occurrences of
```
\adddotspace,
```
that would leave us with the output
> . ,

with a space between the dot and the comma

I guess it should be
```
\adddot,
```
or even easier and in this case equivalent
```
.,
```

kristiank:
I will read up on Git usage. But is this a problem to get these edits squashed in? I think further work on Estonian will now be more sporadic when I get hold of more information/terminology.

kristiank:
I will look at the dots and spaces and push a commit.

kristiank:
Is there a way to control whether numerals are with numbers or with roman numerals?

moewew:
Re the Roman numerals: Depends what we are talking about. In principle ~everything~ quite a bit is possible, but not everything is desirable.

Do you mean? Things like page numbers or `volume` numbers? I *really* prefer not to change this in the localisation file. This should be a style decision and `.lbx` files should try to stay clear of style decisions as far as possible (I know we haven't always managed to do that, just look at `magyar.lbx` and `french.lbx`, but it would need a really good reason). If you are talking about the date (month numbers) then this might be within the scope of the `.lbx` file.

moewew:
As for the commit mess: When all your local changes have been merged (and you have double checked that nothing would be lost when you kill your local `dev` branch) you can overwrite your local `dev` with the `dev` here as described in https://stackoverflow.com/q/9210446.

kristiank:
Thank you. I will look into other solutions as well, just to get more substantial with my Git practices.

moewew:
Oh one more thing: Does Estonian ever need declensions for month names in dates? The original file was based on `finnish.lbx` which apparently needed this (helmikuu 2018 vs 1. helmikuuta 2018). I'm asking because the translations for `january` and `basicjanuary` are the same.

Can you just give me a short rundown of how Estonian dates work? So how would you say the date 2018-08-07 in short (like say 7. 8. 2018 in German or 8/7/2018 in American English) and in long form (7. August 2018 or August 7th, 2018 in American)? How would you format long/short date ranges like `2018-07-09/2018-07-11` (e.g. 9.-11. 7. 2018), `2018-06-08/2018-08-09`, `2018-04-03/2019-01-06`. Where do you place the BC/AD markers? What about "circa" markers?

So what would you expect from
```
\documentclass[estonian]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=authoryear, backend=biber, mergedate=false, date=terse]{biblatex}
%short, long, terse, comp

\usepackage{filecontents}
@book{appleby,
  author  = {Humphrey Appleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018-08-07},
}
@book{bppleby,
  author  = {Humphrey Bppleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018-07-09/2018-07-11},
}
@book{cppleby,
  author  = {Humphrey Cppleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018-06-08/2018-08-09},
}
@book{dppleby,
  author  = {Humphrey Dppleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018-04-03/2019-01-06},
}
\end{filecontents}

\addbibresource{\jobname.bib}

\begin{document}
\nocite{*}
\printbibliography
\end{document}
```
with the options `date=long`, `date=short`, `date=terse`, `date=comp`. Is everything looking alright or is there something off?

moewew:
Merged. Discussion for Estonian dates can happen at #801.

Issue #752, opened 2018-05-30
[PATCH] authoryear.bbx: add a macro to tell whether a separate date is to be formatted


CasperVector:
As the title.  This helps to simplify boilerplates like [this](https://github.com/CasperVector/biblatex-caspervector/blob/c837e1491ebe318e7c146462ac0f336feb87437c/tex/caspervector-ay.bbx#L30); the `bbx:ifnodate` macro can be used like [this](https://github.com/CasperVector/biblatex-caspervector/blob/c837e1491ebe318e7c146462ac0f336feb87437c/tex/blx-caspervector-base.def#L157).
```diff
diff --git a/tex/latex/biblatex/bbx/authoryear.bbx b/tex/latex/biblatex/bbx/authoryear.bbx
index 0a2a8a74..8424d40f 100644
--- a/tex/latex/biblatex/bbx/authoryear.bbx
+++ b/tex/latex/biblatex/bbx/authoryear.bbx
@@ -25,6 +25,8 @@
   \usebibmacro{date+extradate}}
 
 \def\bbx@opt@mergedate@true{\bbx@opt@mergedate@compact}
+\newbibmacro*{bbx:ifnodate}[2]{#2}
+\renewbibmacro*{date}{\usebibmacro{bbx:ifnodate}{}{\printdate}}
 
 % merge date/issue with date label
 \def\bbx@opt@mergedate@maximum{%
@@ -37,10 +39,9 @@
             \setunit*{\addspace}%
             \printdateextra}
            {\printlabeldateextra}}}}%
-  \renewbibmacro*{date}{%
+  \renewbibmacro*{bbx:ifnodate}[2]{%
     \iflabeldateisdate
-      {}
-      {\printdate}}%
+      {#1}{#2}}%
   \renewbibmacro*{issue+date}{%
     \iflabeldateisdate
       {}
@@ -58,10 +59,9 @@
          \iflabeldateisdate
            {\printdateextra}
            {\printlabeldateextra}}}}%
-  \renewbibmacro*{date}{%
+  \renewbibmacro*{bbx:ifnodate}[2]{%
     \iflabeldateisdate
-      {}
-      {\printdate}}%
+      {#1}{#2}}%
   \renewbibmacro*{issue+date}{%
     \ifboolexpr{not test {\iffieldundef{issue}}
                 or not test {\iflabeldateisdate}}
@@ -80,14 +80,13 @@
     \iffieldundef{labelyear}
       {}
[...]

Issue #801, opened 2018-09-21
Estonian dates


moewew:
Cf #800. `estonain.lbx` is ultimately based on `finnish.lbx`. Apparently, in Finnish declension of the month name is crucial for the proper date format. That's why there are two strings for each month (`basicfebruary` and `february`). In `estonain.lbx` these month strings seem to coincide, which would make them (and the additional code handling them) useless. 

Please double-check the output of the following example
```
\documentclass[estonian]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=authoryear, backend=biber, mergedate=false, date=short, datecirca=true]{biblatex}
%short, long, terse, comp

\usepackage{filecontents}
\begin{filecontents}{\jobname.bib}
@book{appleby,
  author  = {Humphrey Appleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018-08-07},
}
@book{bppleby,
  author  = {Humphrey Bppleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018-07-09/2018-07-11},
}
@book{cppleby,
  author  = {Humphrey Cppleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018-06-08/2018-08-09},
}
@book{dppleby,
  author  = {Humphrey Dppleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018-04-03/2019-01-06},
}
@book{eppleby,
  author  = {Humphrey Eppleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018-02},
}
@book{Fppleby,
  author  = {Humphrey Fppleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018~},
}
@book{gppleby,
  author  = {Humphrey Gppleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018-02/2018-05},
}
@book{hppleby,
  author  = {Humphrey Hppleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018-02/2019-05},
}
@book{ippleby,
  author  = {Humphrey Ippleby},
  title   = {On the Importance of the Civil Service},
  date    = {2018/2019},
}
\end{filecontents}

\addbibresource{\jobname.bib}

\begin{document}
\nocite{*}
\printbibliography
\end{document}
```

For all values of ` date=short`, `long`, `terse`, `comp`.

[...]

Issue #793, opened 2018-08-31
Empty parentheses with missing year


jspitz:
In the MWE below, I'd expect that the parentheses holding issue and year will not be printed if both year and issue are undefined. Instead, empty parentheses are output, as in:

> Doe, Jane and John Due (n.d.). “Example title”. In: _Topnotch Studies_ 4.1 (), pp. 1–13.

I think this is a regression to earlier version, but I cannot put my finger on the version where it changed.

```
\documentclass{article}
\usepackage[english]{babel}
\usepackage{filecontents}
\usepackage[bibstyle=authoryear]{biblatex}

\addbibresource{\jobname.bib}
 
\begin{filecontents}{\jobname.bib}
    @article{Doe.2016,
 		author = {Doe, Jane and Due, John},
 		title = {Example title},
 		pages = {1--13},
 		volume = {4},
 		number = {1},
 		journal = {Topnotch Studies}
    }
\end{filecontents}
 
\begin{document}
\nocite{Doe.2016}
\printbibliography
\end{document}
```

moewew:
I suspect the `\iflabeldateisdate` doesn't correctly pick up the case where `labeldate` is `nodate`.

However, the MWE has always yielded bad results with non-`authoryear` styles anyway, so I would argue that in this case the input is wrong (`year` or `date` is a required field for `@article` and a published journal article should have a date associated with it. Sure, we are probably talking about accepted publications with no fixed date yet ...).

> Doe, Jane and John Due. “Example title”. In: Topnotch Studies 4.1 (), pp. 1–13.

jspitz:
But see the linked ticket for `biblatex-publist`. In this package, we output the year earlier and then clear it via `\clearfield`, so that the year is not output twice (note that `biblatex-publist` tries to be as bbx-agnostic as it possibly can). This does not work (any longer) with `labeldateparts=false` (it still works with `labeldateparts=true`). Investigating that revealed that there is something fishy in general. The testcase is just the simplest one I could come up with. I agree that it is a rather heuristic one.

jspitz:
And again, I am pretty sure this is a regression in newer versions of biblatex.

moewew:
I should have been clearer, sorry.

[...]

Issue #754, opened 2018-06-04
Wrapper formats for lists and name lists


moewew:
At the moment `\DeclareListFormat` and `\DeclareNameFormat` are applied to each (name) list item separately, this allows for per-item formatting, but not for an overall (name) list wrapper. If one wants to format the entire list one could either format each item and hope that the result is roughly as expected or execute a command at the start of the first entry and at the end of the last entry. But that does not really feel right...

So `\DeclareOuterListFormat`/`\DeclareOuterNameFormat` (suggestions for better names are very much appreciated, theoretically we could also recycle `\DeclareFieldFormat`, but that could cause confusion or backwards compatibility issues in weird set-ups) might come in handy. I had a short look at the code and I think it should be possible to implement a wrapper without undue difficulties.

**Edit** First, simple solution using `\DeclareFieldFormat` is https://github.com/moewew/biblatex/commit/240944da581f102467e2ba2f18a47b94ad303f18 at https://github.com/plk/biblatex/compare/dev...moewew:listwrapper

I can think of one or two very niche uses for this, so the question is whether or not this would be considered interesting enough for the general public to implement.

plk:
I can see the use of this. It would be nice to rename the existing list macro to `\DeclareListItemFormat` and use `\DeclareListFormat` for the new functionality, with suitable warning. Names are easier as we can leave the current one alone and use `\DeclareNamesFormat` for the new functionality.

moewew:
I like `\DeclareNamesFormat`, even though the **`s`** is easy to miss. But I'm wary about changing the meaning of fairly high-level macros like `\DeclareListFormat`. Even though I assume it is not up there with the likes of  `\DeclareFieldFormat` and `\DeclareNameFormat` I fear it could break too many people's documents.

moewew:
A new attempt is at https://github.com/plk/biblatex/compare/dev...moewew:listwrapper. I decided that *wrapper* as in `NameWrapper` and `ListWrapper` would be a natural name for this feature.

I'll prepare a pull request tomorrow if I don't find anything wrong with the commit.

PR #810, opened 2018-09-30
Implement test for merged date (#752)

moewew:
If `bbx:ifmergeddate` is true, the date has already been printed and need not be printed again.

The `date` bibmacro can be moved out of the `mergedate` definitions now, `issue+date` is still defined within the `mergedate` setup.

Cf. #752.

### DIFF
--- a/doc/latex/biblatex/CHANGES.md
+++ b/doc/latex/biblatex/CHANGES.md
@@ -37,6 +37,19 @@
   `nametitledelim`.
   For compatibility reasons `\labelnamepunct` still pops up in the code here
   and there, but `nametitledelim` should be preferred now.
+- `authoryear.bbx` now has a macro `bbx:ifmergeddate` that can be used to
+  check whether the date has been printed at the beginning of an entry
+  and can thus be suppressed later in the `date` and `issue+date` macros.
+  The macro works like a test and expands to the `<true>` branch if the date
+  has been merged and can be suppressed, it expands to `<false>` if the date
+  has not been merged. In practice `\printdate` should then be issued
+  if and only if the test yields false.
+
+  This change means that the definition of the `date` macro can be the same for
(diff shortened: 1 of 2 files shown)


plk:
Merge and I'll run the tests

plk:
Tests are fine

moewew:
Very good. Thanks. I merged another two pull requests.

There is still #816 I need to think about and a few more issues (namely #733 where I have something that is almost OK, #754 where I would like to get something done with good command names, #793 where I am still in principle opposed to doing something, but where we probably *should* do something to make things prettier for users). Given #821 and the fact that the last release was more than half a year ago and lots of stuff changed I would like to think about a new release in the not-too-distant future. Is there anything from your side or from the Biber side that needs looking at before we can move?

moewew:
Very well. I'll tell you when I'm ready. It is probably a good idea to send a short announcement to comp.text.tex and notify the affected package authors for #816, so I will do that then as well. I would like to allow a bit of time for testing and feedback by others and the time needed by Biber binary builders would be just fine for that I think.

moewew:
Merged https://github.com/plk/biblatex/pull/829 and https://github.com/plk/biblatex/pull/830. It would be great if you could run the tests and upload a dev version to SourceForge.

I'm planning on sending out an announcement to c.t.t. asking people for feedback and testing very soon. 

Still waiting to hear back from style developers on the `xstring` issues (https://github.com/plk/biblatex/pull/816), I'm also waiting at least until the weekend to see what to do about https://github.com/plk/biblatex/issues/801. I'm also still torn on #793.

So my current plans are to ask you to start building Biber at the weekend, but I might have to wait a bit longer in case something comes up.

plk:
All tests ok and uploaded.

moewew:
Thank you. Sent a message to c.t.t.

PR #811, opened 2018-10-05
Enhance reprinfrom for articles

moewew:
Journal publication data should also be included here.

This would change the output in people's documents, but I can't really see this impacting a lot of people negatively. If merged, some tests may fail since the output for `moore:related` will change (for the better, I'd say)
```
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\begin{document}
\cite{moore:related}
\printbibliography
\end{document}
```

### DIFF
--- a/tex/latex/biblatex/biblatex.def
+++ b/tex/latex/biblatex/biblatex.def
@@ -2966,7 +2966,9 @@
        \usebibmacro{publisher+location+date}%
        \newunit\newblock
        \usebibmacro{chapter+pages}}
-      {\printfield[title:hook]{journaltitle}%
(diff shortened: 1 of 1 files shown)


plk:
Seems fine to me to merge

moewew:
Good. I'll merge then. Could you then run the test suite and upload changed PDF files if that should be necessary, please?

plk:
Will do.

moewew:
Thank you.

Merged.

plk:
Tests were ok apart form the added journal related info in one entry as expected so I updated the test docs and uploaded the latest.

Issue #738, opened 2018-04-21
labelprefix=@ causes undefined references


moewew:
See https://github.com/plk/biber/issues/223

```latex
\documentclass[british]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=numeric, backend=biber, defernumbers=true]{biblatex}
\addbibresource{biblatex-examples.bib}

\begin{document}
\cite{sigfridsson,ctan}
\printbibheading
\printbibliography[heading=subbibliography,title={Not internet},nottype=online]
\newrefcontext[labelprefix=@]
\printbibliography[heading=subbibliography,title={Internet},type=online]
\end{document}
```
gives undefined references and nulled-out citation numbers. 

Adding `\makeatletter` as in `\makeatletter\newrefcontext[labelprefix=@]\makeatother` fixes the problem, so this probably has to do with category codes and probably `detokenize`.

moewew:
The problem here is that the labelprefix (`\blx@refcontext@labelprefix`) is part of the internal naming structure which should ideally use only safe characters while at the same time appears in the output where much more should be possible.

@plk Maybe we should completely separate these two functions of `labelprefix` into an internal designator `\blx@refcontext@labelprefix` and the actual output `\abx@field@labelprefix` more aggressively. 
As far as I can see most of the separation is already there: We just need to make sure that
- `\blx@refcontext@labelprefix` only contains safe characters: We could do this by using a counter value.
- Pass on both `\blx@refcontext@labelprefix` and the actual desired `labelprefix` value to Biber in the `.bcf`. Use `\blx@refcontext@labelprefix` to construct the refcontext name, use `labelprefix` to populate `\field{labelprefix}` in the `.bbl`.

moewew:
As it turns out a simple counter is probably not the right way to generate the internal `\blx@refcontext@labelprefix`: We want to be able to re-use refcontexts.

moewew:
@plk Do you have any idea what we could do here? At the moment `labelprefix` is expanded and used as part of an internal designator for the bibliography context. That strongly restricts the possible values it may have and can lead to unexpected behaviour as shown in the MWE. Ideally we would implement a way to use arbitrary text as `labelprefix` and use an internal safe designator for the refcontext.

plk:
[...]

PR #816, opened 2018-10-07
Drop xstring

moewew:
Following https://github.com/plk/biblatex/pull/813 I investigated whether it would be possible to drop `xstring` altogether. It turned out that `biblatex` was only using very few commands in very specific places and that it was possible to come up with more primitive replacements there.

It also turned out that `pdftexcmds` was only used in one macro and that that could also be replaced with `etoolbox` macros.

This is more of a code aesthetic suggestion rather than one that resolves unwanted behaviour. So I'd like to discuss this first before merging.

Currently at least the following styles use `xstring` macros
- [ ] biblatex-abnt (https://github.com/abntex/biblatex-abnt/issues)
- [x] biblatex-archaeology (https://github.com/CarlOrff/biblatex-archaeology/issues)
- [x] biblatex-bath (https://github.com/alex-ball/bathbib/issues)
- [ ] biblatex-gb7714-2015 (loads `xstring`, but with a comment that suggests that this is only done to support older versions) (https://github.com/hushidong/biblatex-gb7714-2015/issues)
- [x] biblatex-oxref (loads `xstring`) (https://github.com/alex-ball/biblatex-oxref/issues)
- [x] biblatex-true-citepages-omit (loads `xstring`)
- [x] biblatex-swiss-legal (loads `xstring`)

If we decide not to load that package any more we would probably have to notify the package authors to explicitly load `xstring` themselves.

What do you think?


josephwright:
Looks like a good plan!

moewew:
@josephwright Cool. Comments on the quality of my suggested replacements and corrections or other suggestions are very welcome.

josephwright:
The replacement for `\pdf@strcmp` is fine provided there's no need for expandable code. I'll check the string decomposition ones.

moewew:
If we want to merge https://github.com/plk/biblatex/compare/dev...moewew:hashlpm for #738 we need to keep `pdftexcmd` for `\pdf@mdfivesum`. With `pdftexcmd` back in the play we could resurrect `\blx@xifstrcmp` in its previous form. I don't really see the point, because `\blx@xifstrcmp` is used only in very specific situations and need not be expandable, but I'm open for a discussion.

`xstring` can still go, though. @josephwright Did you find any obvious blunders in my poor man's `xstring` replacements?

moewew:
@plk Can we get this into 3.12?

plk:
Yes, just push it and I'll re-upload. I only count it as a "release" when it hits CTAN and I won't push that until the biber binaries are all ready which will be a week or so.

moewew:
Ah OK. Very good.

I'll send out another announcement to c.t.t. Not that anyone got back to us after the last one, but you never know.

PR #828, opened 2018-10-16
Fix typo

dbitouze:
BTW, it would be worth consider using `cleveref`'s `\cref` instead of `\secref`, `\Secref`, `\apxref`, `\Apxref`, `\tabref`, `\Tabref`: this unique command would insert the right term (\S -- needs some configuration--, "appendix", "table") depending on the object which is referenced and would save the doc's author from worrying about the type of the object they want to reference.

### DIFF
--- a/doc/latex/biblatex/biblatex.tex
+++ b/doc/latex/biblatex/biblatex.tex
@@ -348,7 +348,7 @@ In that case you need to make sure that all redefined macros used by \biblatex\ 
 \biber\ versions are closely coupled with \biblatex\ versions. You
 need to have the right combination of the two. \biber\ will throw a fatal error
 during processing if it encounters information which comes from a
-\biblatex\ version which is incompatible. \tabref{tab:int:pre:bibercompat} shows a
+\biblatex\ version which is incompatible. \Tabref{tab:int:pre:bibercompat} shows a
 compatibility matrix for the recent versions.
 
 \begin{table}


moewew:
Thank you. Merged.

If I were to start a new document from scratch now I would probably use `cleveref`, but I don't know if it is really worth switching over an existing document to `cleveref`. The most common type of reference is `\secref` and I sort of got used to it.

That said, if you are willing to rework the documentation to use `cleveref` and check that the output is as expected I'd be happy to merge a pull request. It just is not on top of my agenda to do that myself.

There are a few other things about the docs as well. They are based on PL's now unmaintained `ltxdockit` class and I noticed some quirks with that class when I wrote the documentation for `biblatex-ext` [[example fix](https://github.com/moewew/biblatex-ext/blob/a4a1ea3cf7beab10eddedc90d79eefb84b2879b9/biblatex-ext.tex#L321)]. For starters, many KOMA options are outdated and throw warnings. And the code loads `hyperref` very early ...

dbitouze:
>If I were to start a new document from scratch now I would probably use cleveref, but I don't know if it is really worth switching over an existing document to cleveref. The most common type of reference is `\secref` and I sort of got used to it.

I could imagine :)

>That said, if you are willing to rework the documentation to use `cleveref` and check that the output is as expected I'd be happy to merge a pull request. It just is not on top of my agenda to do that myself.

I just tried to compile the doc but got errors, notably because of Courier New not installed on my PC. But never mind: it just is not on top of my agenda either ;)

PR #841, opened 2018-11-11
Add missing languages in Latvian translations.

rhssk:


moewew:
Merged. Thank you very much

Issue #863, opened 2019-01-25
More control over date formats


gusbrs:
Currently, date formats (year, short, long, terse etc.) are preamble only options which affect any given datetype globally in the document. It would be interesting, from an user’s perspective, to be able to have more flexibility than that.

Consider two possibilities:

- One would like to have different date formats, for the same datetype, for different entrytypes.
- One would like to have different date formats, for the same datetype and entrytype, in their short and long format citations.

(Conceptually, the situation could be extended to time too. Though I see less relevance there.)

As it stands, the only way around this, that I know of, is to extend the datamodel and create, e.g. `shortdate`, use a sourcemap to copy the contents of `date` to `shortdate` and, then, use `\printdate` or `\printshortdate` as required.

I envisage two (non-excluding?) possible interfaces for this. One would be to add an optional argument to the `\printdate`s commands, as in e.g. `\printdate[long]`. The other, more interesting, I think, but probably more complicated to implement, is to incorporate the date format as an additional optional argument in the formatting directives for date fields. Or provide an equivalent mechanism for the date format in particular (`\DeclareDateFormat`?).

I’m well aware date fields are “special” (and tricky). I make the request from an user’s perspective, not knowing if this might be technically viable. I’ve argued though the reasons I think this could be  desirable.

Below, I leave two MWEs which may serve as test bases for each of the use cases mentioned above.

MWE1:

```
\documentclass{article}

\usepackage{filecontents}
\begin{filecontents}{\jobname.bib}
  @book{book1,
    author = {Author},
    title = {Book title},
    date = {2011-03-04},
  }

  @article{article1,
    author = {Buthor},
    title = {Article title},
    journal = {Foo},
    date = {2003-12-17},
  }
\end{filecontents}

\usepackage[style=authoryear]{biblatex}

\addbibresource{\jobname.bib}

% could this date also be ‘short’, besides ‘bold’?
\DeclareFieldFormat[article]{date}{\mkbibbold{#1}}

\begin{document}

\nocite{*}

\printbibliography

\end{document}
```

MWE2:

```
\documentclass{article}

\usepackage{filecontents}
\begin{filecontents}{\jobname.bib}
  @customa{custom1,
    author = {Author},
    title = {Custom long title longer than the short one},
    shorttitle = {Short title},
    date = {2011-03-04},
  }
\end{filecontents}

\usepackage[style=verbose]{biblatex}

\addbibresource{\jobname.bib}

\DeclareBibliographyDriver{customa}{%
[...]

PR #866, opened 2019-01-29
Consolidate option declarations

moewew:
Introduce a new macro `\DeclareBiblatexOption` that can be used to declare options at several scopes/levels at once to help avoid code repetition.

Use that new macro to streamline option definitions and enable a few options in additional scopes.

Related to #863, but only in spirit and not for the substance of #863.


plk:
I need to know which option scopes this changes as there need to be `.bcf` schema changes and also `biber` code changes to match. Basically, any changes to the option scope table in the `biblatex` doc need `biber` changes. It is also possible that some option scopes are not possible depending on the code. Would it be easy to prepare a list of option scope changes so I can investigate?

moewew:
I think I only changed options that did not involve *active* work for Biber, the changed options are

### Scopes added (global to global, per-type, per-entry)
- `abbreviate`
- `citetracker`
- `clearlang`
- `dateabbrev`
- `ibidtracker`
- `idemtracker`
- `loccittracker`
- `opcittracker`

### Scopes added (global to global, per-type, per-entry, per-namelist, per-name)
- `$<$namepart$>$inits`
- `terseinits`

### Datatype changed from `boolean` to `string`

- `hyperref`
- `pagetracker`
- `citecounter`
- `citetracker`
- `ibidtracker`
- `idemtracker`
- `opcittracker`
- `loccittracker`

since they take more values than true and false.


### Candidates for extension where work on Biber-side is likely to be necessary
I would have liked to look into more scopes for

- `labelalpha` (add per-entry)
- `labeltitle` (add per-entry)
- `labeltitleyear`  (add per-entry)
- `labeldateparts` (add per-entry)
- `labelnumber` (add per-entry)
- `singletitle` (add per-entry)
- `uniquetitle` (add per-entry)
- `uniquebaretitle` (add per-entry)
- `uniquework` (add per-entry)
- `uniqueprimaryauthor` (?)
- `uniquename` (per-name? and per-namelist?)
- `uniquelist` (per-namelist?)
- `skipbib` (global, granted that does not make sense if there are no `skipbib=false`s on per-type or per-entry level to cancel that out, so that only if is easy on the Biber side)
- `skipbiblist` (global, ditto)
- `skiplab` (global, ditto)
- `dataonly` (global, ditto)

but I believe that those would require changes to Biber.

Ultimately I did this so I could look at per-type date options, but there are other difficulties there are the moment, so I will probably look at that later.

plk:
I think that this is all fine apart from the last 4 - I can't really see any use of these options globally as that would be tantamount to not really using them at all. These options are only really ever used internally when `biber` creates clones when processing related entries, sourcemap clones etc.

moewew:
I agree that adding 'global' for the last four does not make sense on its own, but I could imagine that one would want to process all but a few types and then one could set `skipbib` globally to true and toggle it to false for, say `@article`s. At the moment it would be possibly, if slightly awkward, to get the same result by looping over all entry types. I don't have any more arguments to convince you, but if it is easy to implement this on the Biber-side, why not go for it?

If you push the required changes for the last batch of options (`labelalpha` etc.) to Biber, it should be easy to add another commit that implements the `biblatex` side of things.

plk:
A draft of the outstanding scope changes is now implemented. I updated bibalatex too as I had to check .bcf validation with the changes - seems there are merge conflicts with your branch now but I think this will be minor as the biblatex changes were very simple. I haven't tested the global `skip*` options much yet. The granular scope for `uniquename` was tricky but seems ok.

plk:
Ah, of course, I did half-register this late late night. Will look at it.

plk:
This should be fixed now. I have changed names to simply `un/ul` for the actual biber output numeric counters and retained the full names for the per-entry biblatex options which are passed to the .bbl (these are in fact only used for some error checking). That wasn't the problem actually - I needed to do a complete tidy and audit of the entry/namelist/name option scope handling, which is now done - there were some missing input/output handlers due to all the options churn over the last few releases.

plk:
I have some more refactoring to do I think - I don't want any more hard-coding of input/output option handling in biber and so I will likely enhance the .bcf with something to make this dynamic.

moewew:
OK, if it is more convenient for you if we hold off merging the changes from here for a while, let me know.

plk:
I have done what I need to do with this now and 2.13/3.13 are updated. The .bcf has been augmented with information about which per-entry and below options to output to the .bbl and what to set when certain meta-options are defined. This was a bit messy and hard-coded in places. Now it is all determined via the .bcf. In the process, I changed things a bit to make what `biber` outputs to the .bbl consistent. Mostly this just means  that the `dataonly` options are broken out explicitly in the .bbl and always passed back, which wasn't always happening before and was in fact a bug. I still haven't played that much with global `dataonly` etc.

moewew:
Squashed and merged the changes from here (in https://github.com/plk/biblatex/commit/9bef9aced8c41cdc568a7f50c34a4be3aee688b4). I've moved some options and their declarations around, but there should be no functional change apart from the things discussed at the beginning here. 

My tests all ran OK, but I only just realised that 95 always fails for me because I don't have the `.bib` file. Can you push `95-customlists.bib` to https://github.com/plk/biblatex/tree/dev/doc/latex/biblatex/examples, please?

plk:
Very nice work on the options consolidation - it is much neater now. I have uploaded up-to-date 2.13/3.13 to SF.

PR #881, opened 2019-04-09
Add `doi` field for `@online` type

moewew:
It seems useful to have a `doi` also for `@online` entries. Especially since the already support the more general `eprint`.

See c53c1f5b29b14e6a5c9d005ddebcdcf5b8b52270, 4c18e0fce96c76f27c21ef898ea918c500faf9af, b66a1d53a825c6631bcbb17475a252b071b84b99.

### DIFF
--- a/doc/latex/biblatex/biblatex.tex
+++ b/doc/latex/biblatex/biblatex.tex
@@ -527,7 +527,7 @@ A fallback type for entries which do not fit into any other category. Use the fi
 
 An online resource. \bibfield{author}, \bibfield{editor}, and \bibfield{year} are omissible in terms of \secref{bib:use:key}. This entry type is intended for sources such as web sites which are intrinsically online resources. Note that all entry types support the \bibfield{url} field. For example, when adding an article from an online journal, it may be preferable to use the \bibtype{article} type and its \bibfield{url} field.
 
-\reqitem{author/editor, title, year/date, eprint/url}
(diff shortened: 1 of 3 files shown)


moewew:
It would be prettier to have the `@online` driver just call
```latex
  \usebibmacro{date}%
  \newunit\newblock
  \usebibmacro{doi+eprint+url}%
  \newunit\newblock
  \usebibmacro{addendum+pubstate}%
```
instead of
```latex
  \usebibmacro{date}%
  \newunit\newblock
  \iftoggle{bbx:doi}
    {\printfield{doi}}
    {}%
  \newunit\newblock
  \iftoggle{bbx:eprint}
    {\usebibmacro{eprint}}
    {}%
  \newunit\newblock
  \usebibmacro{url+urldate}%
  \newunit\newblock
  \usebibmacro{addendum+pubstate}%
```
But there are two small issues with that:
1. `@online` disobeys the `url` option and always shows the URL. If it were to use `doi+eprint+url` it would check the toggle `bbx:url`. Since it is possible in 3.13 to set the `url` option per type it should be possible to circumvent this issue.
2. There might be backwards incompatibility issues with code that either patches the `@online` driver or that modifies `doi+eprint+url`, but assumes that it does not apply to `@online`. I have my doubts that there is going to be a lot of trouble because of those assumptions, but you never know.

Comments and thoughts are appreciated.

Issue #875, opened 2019-03-01
`<option>' undefined in families `blx@opt@typ'


pcdi:
If I try ```\ExecuteBibliographyOptions[book]{eprint=true}``` as per §[3.2.2](http://mirror.ctan.org/macros/latex/contrib/biblatex/doc/biblatex.pdf#subsubsection.3.2.2), I get ```! Package keyval Error: eprint undefined.``` or ```Package xkeyval Error: `eprint' undefined in families `blx@opt@typ'.```

However,
> [**3.1.2.2 Style-specific**](http://mirror.ctan.org/macros/latex/contrib/biblatex/doc/biblatex.pdf#paragraph.3.1.2.2) […] Technically, they are preamble options like those in §[3.1.2.1.](http://mirror.ctan.org/macros/latex/contrib/biblatex/doc/biblatex.pdf#paragraph.3.1.2.1)

so I think ```eprint``` and the other style-specific options mentioned in the documentation should be included in ```blx@opt@typ``` to also behave the same way as the general options.

moewew:
In light of https://github.com/plk/biblatex/pull/866 that would make sense.

The only issue would be if there are custom styles that already say `\DeclareTypeOption[boolean]{eprint}{...}`, because that would clash with our new definitions. A short search in the custom styles installed on my system showed no problematic definition, so we can probably go ahead.

M(n)WE
```latex
\documentclass[british]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=authoryear, backend=biber]{biblatex}

\ExecuteBibliographyOptions[book]{eprint=false}

\addbibresource{biblatex-examples.bib}

\begin{document}
\cite{baez/online,wilde}
\printbibliography
\end{document}

```

moewew:
See https://github.com/plk/biblatex/compare/dev...moewew:styleopts. Currently the branch contains those options that were straightforward.

It wasn't quite clear to me what should happen to `dashed` implementation-wise and output-wise, so I left it as is for the moment. Same for the citation tracking options `citepages`, `ibidpage`, `strict` and `pageref`.

moewew:
https://github.com/plk/biblatex/pull/877 has been merged. Most style options that make sense in non-global scope should be available per-type and per-entry in the next `biblatex` release.

moewew:
[...]

Issue #857, opened 2019-01-18
refsection=chapter does not work with KOMA-Script 3.26a


komascript:
I've got a [bug report about an issue using `refsection=chapter` with KOMA-Script 3.26a](https://komascript.de/node/2220). With this example:

    \listfiles
    \documentclass{scrbook}
    \usepackage[backend=biber, refsection=chapter]{biblatex}
    \begin{filecontents}{\jobname.bib}
      @misc{B1,
        title = {Bib1}
      }
      @misc{B2,
        title = {Bib2}
      }
    \end{filecontents}
    \addbibresource{\jobname.bib}

    \begin{document}
    \chapter{Ch1}
    \nocite{B1}
    \printbibliography
    \chapter{Ch2}
    \nocite{B2}
    \printbibliography
    \end{document}

Both chapters show the complete bibliography not only the cites of the chapter. A look into the `aux`-file (and comparing it with the same file using `book` instead of `scrbook`) shows, that `refsection` is not increased and the `\newlabel` commands are missing. I do not know enough about the `biblatex` code to find out the reason. I only found comments about code changes for KOMA-Script 3.26.

Please have a look into it. If I can help to fix the issue, e.g. adding another hook, please tell me.

moewew:
Thank you for reporting this.

This is probably related to https://github.com/plk/biblatex/issues/713 and https://komascript.de/node/2168. I (wrongly, as it now turns out) assumed that I could also use the new `\At@startsection` to patch `\chapter` as well. The earlier code (`\pretocmd\@makechapterhead` and `\pretocmd\@makeschapterhead`) seems to do the right thing, so we could go back to that. But if there is an official hook to patch into chapter headings I would of course prefer to use that.

komascript:
I see.

`\chapter` (and `\part`) do not use `\scr@startsection`. So (for me) it is clear, that the hook of `\At@startsection` is not executed. Before I add corresponding commands `\At@startchapter` and `\At@startpart` and hooks for chapter and part style commands, I have to have a look into the code, whether or not there are already hooks, because there are a lot of hooks in the headings mechanism, because several wrapper class and package authors asked for hooks. I really cannot keep all of them in mind. :(

But currently I need some time.

e-kwsm:
Just in case, the version of biblatex is 2018/11/02 v3.12 (sorry for cross post).

moewew:
[...]

PR #887, opened 2019-04-28
Use KOMA-Script hooks (#875)

moewew:
Start using KOMA-Script hooks for section patching. Also fixes an error in the patching code for KOMA-Script versions without the new hook mechanism

https://github.com/plk/biblatex/issues/857

*Merge this when KOMA-Script v3.27 with the hooks is released or just before `biblatex` 3.13 gets ready.*

### DIFF
--- a/tex/latex/biblatex/biblatex.sty
+++ b/tex/latex/biblatex/biblatex.sty
@@ -9833,6 +9833,33 @@
   }%
 }
 
+%% Patching all the different document classes and their sectioning
+%% commands is ... messy at best
+%% For now we do the big three: (1) Standard classes, (2) KOMA-Script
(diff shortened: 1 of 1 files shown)


moewew:
Turns out the exact timing of the patch was wrong for other document classes as well. https://tex.stackexchange.com/q/488547/35864 shows an issue with `memoir`, but the same issue can be reproduced with the standard classes. Fortunately, `memoir` also has heading hooks, so we can use them (https://github.com/plk/biblatex/pull/887/commits/9a617ad88d6dd6cf84cad6d9e2b5216622b668ed). The standard styles don't have hooks and probably need a more precise `\patchcmd` instead of `\pretocmd`... Still to do.

Issue #853, opened 2019-01-06
\mkbibcompletename: A formatting command for the entire name


moewew:
It might be useful to have a command like `\mkbibcompletename` that can be used to format the entire name of a person at once. The piecemeal approach with `\mkbibnamefamily`, `\mkbibnamegiven` is not always enough, especially in `family-given` situations when a comma is involved (and even if it gives acceptable output, it might feel not entirely right).

```latex
\documentclass[british]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=authoryear, backend=biber]{biblatex}

\addbibresource{biblatex-examples.bib}

\DeclareNameAlias{sortname}{family-given}


%%%% current approach
\renewcommand*{\mkbibnamefamily}{\textbf}
\renewcommand*{\mkbibnamegiven}{\textbf}
\renewcommand*{\mkbibnameprefix}{\textbf}
\renewcommand*{\mkbibnamesuffix}{\textbf}

%%%% "\mkbibnamecomplete"
\newcommand*{\mkbibnamecomplete}{\textbf}

\renewbibmacro*{name:family}[4]{%
  \ifuseprefix
    {\usebibmacro{name:delim}{#3#1}%
     \usebibmacro{name:hook}{#3#1}%
     \mkbibnamecomplete{%
       \ifdefvoid{#3}
         {}
         {\ifcapital
            {\mkbibnameprefix{\MakeCapital{#3}}\isdot}
            {\mkbibnameprefix{#3}\isdot}%
          \ifprefchar{}{\bibnamedelimc}}%
          \mkbibnamefamily{#1}\isdot}}
    {\usebibmacro{name:delim}{#1}%
     \usebibmacro{name:hook}{#1}%
     \mkbibnamecomplete{%
       \mkbibnamefamily{#1}\isdot}}}%

\renewbibmacro*{name:given-family}[4]{%
  \usebibmacro{name:delim}{#2#3#1}%
  \usebibmacro{name:hook}{#2#3#1}%
  \mkbibnamecomplete{%
    \ifdefvoid{#2}{}{\mkbibnamegiven{#2}\isdot\bibnamedelimd}%
    \ifdefvoid{#3}{}{%
      \mkbibnameprefix{#3}\isdot
      \ifprefchar
        {}
        {\ifuseprefix{\bibnamedelimc}{\bibnamedelimd}}}%
[...]

PR #888, opened 2019-04-28
Add \mkbibcompletename formatting wrapper (#853)

moewew:
`\mkbibcompletename` is actually a meta-command.
The name formats use `\mkbibcompletename<formatorder>`.

As mentioned in #853 the arguments of these macros are more complex, so underlining or case changing macros might be a bit tricky here. Most other font operations will be fine, though.

### DIFF
--- a/doc/latex/biblatex/CHANGES.md
+++ b/doc/latex/biblatex/CHANGES.md
@@ -37,6 +37,11 @@
 - Add the special fields `volcitevolume` and `volcitepages` for finer control
   over the `\volcite` postnote.
 - Add `\AtVolcite` hook to initialise `\volcite` commands.
+- Add `\mkbibcompletename` as well as `\mkbibcompletename<formatorder>`
+  to format complete names.
+  The commands are analogous to `\mkbibname<namepart>` but apply to
+  the entire name printed in format order `<formatorder>`.
+  By default the predefined macros all expand to `\mkbibcompletename`.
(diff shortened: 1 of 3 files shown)


moewew:
@plk Any comments about this? Does the name `\mkbibcompletename` make sense? Are the modifications to a commonly used macro worth it? (I doubt people patch the macros, so I expect little compatibility issues, but still ...)

plk:
What about plain `\mkbibname` since the name parts macros will alway be longer than this? I can see utility in this.

moewew:
As it stands now I implement a separate `\mkbibcompletename<nameformat>` for each `name:<nameformat>` bibmacro, so if we take `\mkbibname<nameformat>` as a basename we get a clash between `name:family` and `\mkbibnamefamily`. That could be avoided by moving the `<nameformat>` bit of the name: `\mk<nameformat>name`, but then we define both `\mkbibfamilyname` and `\mkbibnamefamily`, which might be confusing.

Is it too much to define a separate `\mkbibcompletename<nameformat>` for each `name:<nameformat>`?

plk:
That seems fine to me since I assume it's auto-created from the datamodel like all the other name format macros.

moewew:
No, the commands are (and have to be) created manually, since they are tied to the `name:...` bibmacros that also have to be defined manually.

plk:
Ah yes, this is only formats so you are just providing format for the default datamodel.

PR #766, opened 2018-06-28
Add sortings where volume comes before year

svenper:
This could be useful for reissues. The new sortings are basically copied from `nyvt` and `anyvt`.

Issue #256, opened 2014-07-18
Some kind of mvperiodical booktype


georgthegreat:
Hello.

I'm still working on my database of historical books.
I've recently run into a problem. I've found 6 volumes of 7-paged (each) journals of early guitar music. They were issued monthly (so it's a periodical of some kind), but without changes in title or anything like this.

So, my [commit](https://github.com/georgthegreat/dancebooks-bibtex/commit/a6b66c543ca5af85d48fa757dbc32b6a53108d66) was just a copypaste of 6 similar entries. I wonder if there is a possibility to add @mvperiodical entry type (I really don't need the date distinction).

Issue #156, opened 2013-09-06
EE field and user fields support


kno10:
Hello,
While I'm a big fan of the biblatex package, it would be nice if user fields were supported better, without having to modify biblatex.def

Of particular interest to me is the EE field. This field is commonly filled e.g. when importing DBLP data. The name seems to be shorthand for "electronic edition".

We have EE data for most entries here. In contrast to the "url" field, this is _optional_ data; we use url when it is part of information necessary for the reference to be meaningful.

To save on space, we don't want urls to be printed always. So if there is a URL field, we want it to be printed, if there is an EE field, we don't want it to be printed - but for generating PDF documents, I'd like to have the reference actually _link_ to the EE.

So this bug is two-fold:
A) make "ee" a supported field, and maybe use it in the default templates for hyperlinking.
B) make it easier to add custom fields, without having to modify biblatex.def
(and without having to modify my .bib file to use non-understandable "usera" fields)

thanks.

Issue #122, opened 2013-06-10
Add field for conference acronym


skoehler:
Consider this bibtex entry:

@INPROCEEDINGS{Koehler2013,
  author = {Sven Köhler},
  title = {Some cool result},
  date = {2013},
  eventtitle = {2nd conference on some matter (CSM13)},
  eventdate = {2013-02-01/2013-02-04},
  venue = {Paris, France},
  series = {LNCS},
  number = {9999},
  publisher = {Springer},
  location = {Berlin / Heidelberg},
  pages = {16-30}
}

With the standard biblatex style, it will be typeset as

[1] Sven Köhler. “Some cool result”. In: 2nd conference on some matter (CSM13) (Paris, France, Feb. 1–4, 2013). LNCS 9999. Berlin / Heidelberg: Springer, 2013, pp. 16–30.

Obviously, the double paranthesis is a bit odd, and in fact the output could very well be
[1] Sven Köhler. “Some cool result”. In: 2nd conference on some matter (CSM13, Paris, France, Feb. 1–4, 2013). LNCS 9999. Berlin / Heidelberg: Springer, 2013, pp. 16–30.

But actually, it should be up to the style (and not the bibtex author) where the acronym is placed. To the best of my knowledge, biblatex has no field for the acronym. Please add one.

PR #907, opened 2019-07-15
Patent date

bastien-roucaries:
Add patents date

### DIFF
--- a/bibtex/bst/biblatex/biblatex.bst
+++ b/bibtex/bst/biblatex/biblatex.bst
@@ -111,6 +111,10 @@ ENTRY {
     pubstate
     reprinttitle
     holder
+    fileddate
+    publisheddate
+    granteddate
+    expireddate
     publisher
     school
     series
@@ -1629,6 +1633,27 @@ FUNCTION {output:dates.2} {
   "urlendday"     swap$ output:write:field
   "urlendmonth"   swap$ output:write:field
   "urlendyear"    swap$ output:write:year
+  "fileddate" fileddate split.date
+  "filedday"        swap$ output:write:field
+  "filedmonth"      swap$ output:write:field
+  "filedyear"       swap$ output:write:field
+  "filedendday"     swap$ output:write:field
+  "filedendmonth"   swap$ output:write:field
+  "filedendyear"    swap$ output:write:year
+  "granteddate" granteddate split.date
+  "grantedday"        swap$ output:write:field
+  "grantedmonth"      swap$ output:write:field
+  "grantedyear"       swap$ output:write:field
+  "grantedendday"     swap$ output:write:field
+  "grantedendmonth"   swap$ output:write:field
+  "grantedendyear"    swap$ output:write:year
+  "expireddate" expireddate split.date
+  "expiredday"        swap$ output:write:field
+  "expiredmonth"      swap$ output:write:field
+  "expiredyear"       swap$ output:write:field
+  "expiredendday"     swap$ output:write:field
+  "expiredendmonth"   swap$ output:write:field
+  "expiredendyear"    swap$ output:write:year
 }
 
 FUNCTION {output:user} {
--- a/tex/latex/biblatex/biblatex.def
+++ b/tex/latex/biblatex/biblatex.def
@@ -1258,6 +1258,7 @@
     \step[fieldsource=primaryclass,  fieldtarget=eprintclass]
     \step[fieldsource=key,           fieldtarget=sortkey]
     \step[fieldsource=pdf,           fieldtarget=file]
+    \step[fieldsource=publisheddate  fieldtarget=date]
   }
 }
 
--- a/tex/latex/biblatex/biblatex.def
+++ b/tex/latex/biblatex/biblatex.def
@@ -2804,6 +2804,30 @@
 
 \newbibmacro*{date}{\printdate}
 
+\DeclareFieldFormat{fileddate}{\bibstring{filed}\space#1}
+\DeclareFieldFormat{granteddate}{\bibstring{granted}\space#1}
+\DeclareFieldFormat{expireddate}{\bibstring{expired}\space#1}
+
+\DeclareFieldFormat{patentdates}{\mkbibparens{#1}}
+\newbibmacro*{patentdates}{%
+  \ifboolexpr{
+      test {\iffieldundef{filedyear}} and
+      test {\iffieldundef{grantedyear}} and
+      test {\iffieldundef{expiredyear}}
+    }
+  {}
+  {%
+      \newunit
+      \printtext[patentdates]{%
+        \printfileddate%
+        \setunit*{\addcomma\space}%
+        \printgranteddate%
+        \setunit*{\addcomma\space}%
+        \printexpireddate%
+      }%
+  }%
+}
+
 \newbibmacro*{url+urldate}{%
   \usebibmacro{url}%
   \iffieldundef{urlyear}
--- a/tex/latex/biblatex/biblatex.sty
+++ b/tex/latex/biblatex/biblatex.sty
@@ -14861,9 +14861,9 @@
   \blx@checkencoding
   \if@filesw
     \blx@ctrlwrite
-    \blx@bblinput
-    \blx@maxsection\z@
   \fi
+  \blx@bblinput
+  \blx@maxsection\z@
   \csuse{abx@preamble}%
   \blx@inf@refsec
   \blx@inf@refseg}
--- a/tex/latex/biblatex/biblatex.sty
+++ b/tex/latex/biblatex/biblatex.sty
@@ -5143,6 +5143,9 @@
   \do{winter}%
   \do{am}%
   \do{pm}%
+  \do{filed}%
+  \do{granted}%
+  \do{expired}%
 }
 
 \newrobustcmd*{\NewBibliographyString}[1]{%
--- a/tex/latex/biblatex/lbx/english.lbx
+++ b/tex/latex/biblatex/lbx/english.lbx
@@ -556,6 +556,9 @@ canadian,australian,newzealand,USenglish,UKenglish}
   winter           = {{Winter}{Win\adddot}},
   am               = {{AM}{AM}},
   pm               = {{PM}{PM}},
+  filed            = {{filed}{filed\adddot}},
+  granted          = {{granted}{grant\adddot}},
+  expired          = {{expired}{exp\adddot}},
 }
 
 \protected\gdef\lbx@us@mkdaterangetrunc@long#1#2{%
--- a/tex/latex/biblatex/lbx/french.lbx
+++ b/tex/latex/biblatex/lbx/french.lbx
@@ -559,6 +559,9 @@
   winter           = {{hiver}{hiver}},
   am               = {{AM}{AM}},
   pm               = {{PM}{PM}},
+  filed            = {{d\'epos\'e}{dep\adddot}},
+  granted          = {{accord\'e}{acc\adddot}},
+  expired          = {{\'echu}{\'echu}},
 }
 
 \gdef\lbx@fr@ordinal#1#2{%
--- a/doc/latex/biblatex/biblatex.tex
+++ b/doc/latex/biblatex/biblatex.tex
@@ -892,13 +892,26 @@ The title of a conference, a symposium, or some other event in \bibtype{proceedi
 
 An annex to the \bibfield{eventtitle} field. Can be used for known event acronyms, for example.
 
+\listitem{expireddate}{date}
+
+The date of expiration for a \bibtype{patent}. Patents usually (with a lot of legal exception depending of countries) expire twenty years after filling. This date could be set in order to document that the
+\bibtype{patent} has expired earlier than expected (usually due to fee not payed).
+
 \fielditem{file}{verbatim}
 
 A local link to a \acr{pdf} or other version of the work. Not used by the standard bibliography styles.
 
+\listitem{fileddate}{date}
+
+The  date of filing for a \bibtype{patent}. \bibfield{fileddate} is usually the beggining of the legal monopole.
+
 \listitem{foreword}{name}
 
 The author(s) of a foreword to the work. If the author of the foreword is identical to the \bibfield{editor} and\slash or \bibfield{translator}, the standard styles will automatically concatenate these fields in the bibliography. See also \bibfield{introduction} and \bibfield{afterword}.
+
+\listitem{granteddate}{date}
+
+The granted date for a \bibtype{patent}.
 
 \listitem{holder}{name}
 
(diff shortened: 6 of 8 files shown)


moewew:
Thank you for the pull request. Unfortunately, I'm a bit short of time at the moment and I'm not sure if I can have a look at this in the next week or so. Even if I find the time to comment on a few things there is still the decision of whether or not to accept the pull request.

In the past, the stance has been rather conservative with respect to "simple" changes to the datamodel like this. It has always been accepted that the standard styles can not cater for everyone's need and that sometimes custom definitions are needed: That's why `biblatex` is easily customisable in the first place. See for example previous discussions in https://github.com/plk/biblatex/issues/122, https://github.com/plk/biblatex/issues/156, https://github.com/plk/biblatex/issues/256, https://github.com/plk/biblatex/pull/766

tex/latex/biblatex/biblatex.def
```diff
@@ -2803,6 +2804,29 @@
 
 \newbibmacro*{date}{\printdate}
 
+\DeclareFieldFormat{patentextradates}{\mkbibparens{#1}}
+\newbibmacro*{patentdate}{
```
moewew: This line needs `%` at the end.
```suggestion
\newbibmacro*{patentdate}{%
```

tex/latex/biblatex/biblatex.def
```diff
@@ -2803,6 +2804,29 @@
...
+  \printdate%
```
moewew: Here the `%` should not be needed.
```suggestion
  \printdate
```
One may want to think about using `\usebibmacro{date}%` here instead.
```suggestion
  \usebibmacro{date}%
```
Or maybe
```latex
  \newunit\newblock
  \usebibmacro{date}%
  \newunit\newblock
  \usebibmacro{patentdate}%
```
in `standard.bbx`.

In any case it would have to be made sure that this plays nicely with `mergedate` from `authoryear.bbx`.

tex/latex/biblatex/biblatex.def
```diff
@@ -2803,6 +2804,29 @@
...
+  {%
```
moewew: The group started here seems superfluous.

tex/latex/biblatex/biblatex.def
```diff
@@ -2803,6 +2804,29 @@
...
+  \ifboolexpr{%
+      test {\iffieldundef{filedyear}} and%
+      test {\iffieldundef{grantedyear}} and%
+      test {\iffieldundef{expiredyear}}%
+    }%
+    {}%
+    {%
+      \newunit%
+      \printtext[patentextradates]{%
+          \iffieldundef{filedyear}{}{\bibstring{filed}\space \printfileddate}%
```
moewew: This could be handled more elegantly with field formats, cf. `urldate` and `\DeclareFieldFormat{urldate}{\mkbibparens{\bibstring{urlseen}\space#1}}`.

tex/latex/biblatex/biblatex.sty
```diff
@@ -5143,6 +5143,9 @@
   \do{winter}%
   \do{am}%
   \do{pm}%
+  \do{filed}
```
moewew: The lines here are missing `%`s at the end
```suggestion
  \do{filed}%
```
etc. etc. (apparently, I can't comment on multiple lines at once...)

tex/latex/biblatex/lbx/french.lbx
```diff
@@ -559,6 +559,9 @@
   winter           = {{hiver}{hiver}},
   am               = {{AM}{AM}},
   pm               = {{PM}{PM}},
+  filed            = {{D\'epos\'e}{Dep\adddot}},
```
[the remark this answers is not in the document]
bastien-roucaries: ok

tex/latex/biblatex/lbx/english.lbx
```diff
@@ -556,6 +556,9 @@ canadian,australian,newzealand,USenglish,UKenglish}
   winter           = {{Winter}{Win\adddot}},
   am               = {{AM}{AM}},
   pm               = {{PM}{PM}},
+  filed            = {{Filed}{Fil\adddot}},
```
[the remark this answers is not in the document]
bastien-roucaries: ok

tex/latex/biblatex/biblatex.def
bastien-roucaries: prefer this, note that maybe could legaldate ? Legislation act and other law have legal date:
law are voted, are enacted, then repelled. So maybe something more generic will help here

tex/latex/biblatex/biblatex.sty
bastien-roucaries: ok

tex/latex/biblatex/lbx/english.lbx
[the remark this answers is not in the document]
bastien-roucaries: ok

tex/latex/biblatex/biblatex.def
bastien-roucaries: I do not understand this one

bastien-roucaries:
> Thank you for the pull request. Unfortunately, I'm a bit short of time at the moment and I'm not sure if I can have a look at the big picture in the next week or so. Even if I find the time to comment on a few things there is still the decision of whether or not to accept the pull request.
> 
> In the past, the stance has been rather conservative with respect to "simple" changes to the datamodel like this. It has always been accepted that the standard styles can not cater for everyone's need and that sometimes custom definitions are needed: That's why `biblatex` is easily customisable in the first place. See for example previous discussions in #122, #156, #256, #766

I understand but for patents they are three date printed on a patent : publication date, filed date and granted date. The three date are important from a legal point of view, and it will improve the quality of bibliographic database

bastien-roucaries:
refreshed with your comment

tex/latex/biblatex/biblatex.def
moewew: I think I would prefer `patentdate` or maybe even `patentdates` over `legaldate` as the name of the bibmacro here. Even though laws and regulations (where `legaldate` is a good name) might be in a similar situation, the macro here is specifically targeting dates that have been named for patents only.

bibtex/bst/biblatex/biblatex.bst
```diff
@@ -111,6 +111,10 @@ ENTRY {
     pubstate
     reprinttitle
     holder
+    fileddate
```
moewew: Some more work is needed to properly handle `date` fields in the `.bst`. Cf. everything that involves `urldate` and its components `urlyear` etc.

tex/latex/biblatex/biblatex.def
```diff
@@ -2803,6 +2804,26 @@
 
 \newbibmacro*{date}{\printdate}
 
+\DeclareFieldFormat{legaldates}{\mkbibparens{#1}}
+\newbibmacro*{legaldates}{
```
moewew: The line is still missing a `%` to protect the line end.
```suggestion
\newbibmacro*{legaldates}{%
```
A few protecting `%` in the following lines are unnecessary. Mostly the ones in the first argument of `\ifboolexpr`, but also the one `{}%` in l. 2814 and the one after `\newunit`.

tex/latex/biblatex/biblatex.def
```diff
@@ -2803,6 +2804,26 @@
...
+  \ifboolexpr{%
+      test {\iffieldundef{filedyear}} and%
+      test {\iffieldundef{grantedyear}} and%
+      test {\iffieldundef{expiredyear}}%
+    }%
+  {}%
+  {%
+      \newunit%
+      \printtext[legaldates]{%
+        \iffieldundef{filedyear}{}{\bibstring{filed}\space \printfileddate}%
```
moewew: I haven't tested it, but something like
```suggestion
        \printfileddate
```
would be enough here if we also define
```latex
\DeclareFieldFormat{fileddate}{\bibstring{filed}\space#1}
```

bibtex/bst/biblatex/biblatex.bst
bastien-roucaries: ok done

tex/latex/biblatex/biblatex.def
bastien-roucaries: work done

tex/latex/biblatex/biblatex.def
bastien-roucaries: ok

bastien-roucaries:
refreshed with your comments

tex/latex/biblatex/lbx/french.lbx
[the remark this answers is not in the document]
dbitouze: `filed            = {{D\'epos\'e}{D\'ep\adddot}},`
would be better than the current:
`filed            = {{D\'epos\'e}{Dep\adddot}},`

moewew:
Thank you for the improvements.

I'm afraid to say that at the moment I'm not convinced this is an addition that is *either (1) encountered time and again or (2) can't be solved without changes to internals* (https://github.com/plk/biblatex/issues/256#issuecomment-49529087). I also couldn't find any style guide that requires that all those date be listed (I only checked a few I could get my hands on).

Do you have any evidence that this is an important feature that people *need*, ideally references to established style guides that want addition of these dates?

bastien-roucaries:
According to chicago manual of style:
15.50 Patents or other documents cited by more than one date Cite patents and other documents that include more than one date as follows (note that the year of issue is repeated to avoid ambiguity).

Christensen, Godtfred Kirk. Toy building set. U.S. Patent 3,597,875 filed November 18, 1968, and issued August 10, 1971.

bastien-roucaries:
@moewew  chicago manual of style was established in USA during the period of submarine patent, so issued and granted date were the same. It is not the case now, and was not the case in Europe, so we have three dates + expiration

bastien-roucaries:
IEEE publication differentiate between filled and granted date see ftp://tug.ctan.org/pub/tex-archive/macros/latex/contrib/IEEEtran/bibtex/IEEEtran_bst_HOWTO.pdf

hushidong:
we now have date origdate eventdate enddate，
can we use them as filed granted issued expired  date?

moewew:
@hushidong We could do that, but it wouldn't be particularly semantically sound. It's not that it is *hard* to implement the new dates, it's more about whether or not it makes sense to change the *standard styles* to include these dates. As far as I can see it has always been the general stance of the `biblatex` maintainers that the standard styles can't cover every eventuality and that it makes sense to be conservative when it comes to changes.

@bastien-roucaries Thanks for the references. I haven't had the chance to check what exactly the CMoS has to say about patents first-hand, but most secondary sources say that several patent dates are required.

The reference to http://mirrors.ctan.org/macros/latex/contrib/IEEEtran/bibtex/IEEEtran_bst_HOWTO.pdf is interesting, because the document explains

> In order to provide full support for both patents and patentapplications,  two  sets  of  date  fields  are  provided.  One  set pertains  to  the  date  the  patent  was  granted  (`day`,  `month`  and `year`) the other pertains to the date the patent application filed  (`dayfiled`,  `monthfiled`  and  `yearfiled`).  There  is  a  slight complication  because  the  IEEE  displays  only  one  date  for references of patents or patent applications. `IEEEtran.bst` looks for  the  presence  of  the  `year`  and  `yearfiled`  files.  If  the  `yearfield` is present, the set pertaining to the date granted is used. Otherwise,  `IEEEtran.bst`  uses  the  set  pertaining  to  the  date filed.

In particular I would expect that every patent (as opposed to a patent *application*) would definitely have a date granted, which means that the date filed is rarely used.

See also https://www.ieee.org/content/dam/ieee-org/ieee/web/org/conferences/style_references_manual.pdf, p. 9 or https://journals.ieeeauthorcenter.ieee.org/your-role-in-article-production/ieee-editorial-style-manual/ and in particular http://journals.ieeeauthorcenter.ieee.org/wp-content/uploads/sites/7/IEEE-Reference-Guide.pdf, p. 11

> NOTE: Use “issued date” if several dates are given.

I'm not sure if the CMoS alone is enough to convince me the standard styles need all these dates.

bastien-roucaries:
@moewew  Patent are not always granted since the 2010's and the WIPO rules. Now patent are publied after two years and are not always granted. IEEE style manual use the old US rule, that publied patent are always granted. It is now not the case.

Issued date is important for computing the normal expiration date that is approximatly 20 years after filling date.

Bluebook recommand other stuff but need issued for application https://legalbeagle.com/5650426-cite-patent-application.html

They are a good summary of US style for patent here:
https://legalbeagle.com/5650426-cite-patent-application.html

Acm need filling date (and it is a huge publisher for computer science)
https://www.acm.org/publications/authors/reference-formatting

pubmed ask to add application number and filling date as optionnal 
https://www.ncbi.nlm.nih.gov/books/NBK7260/

Thanks

moewew:
I'm still not entirely sure what to do with this. I'm hoping to finish off `biblatex` 3.13 soon, so I think I will leave this open to discuss and pick up after the 3.13 release for `biblatex` 3.14.

bastien-roucaries:
@moewew  I volonteer for next revision. I can port if needed

moewew:
Oh right, this is still open. Sorry about that.

Since we already have some data model changes lined up for the next version we might as well look at this again. I'll take a closer look again, but I can't promise it will be this weekend, I'm afraid. If there is anything that needs to be changed, I'll let you know.

bastien-roucaries:
Thanks

moewew:
Hmmm, I still have my reservations about this. But they are mostly aesthetic: It just seems to be a lot of additional weight for a very, very narrow use case. (In particular the required change for the `.bst` file and the data model.)

@plk What do you think?

plk:
I think I would opt for a style to implement this as it is quite specialised and I can't see this being used much, if at all, in standard styles. A style could easily add datamodel components to make this possible.

moewew:
That's more or less what I feel as well, but there is an argument to be made that support in the standard styles would help with adoption of these kind of dates in other styles (possibly styles that should or at least could have them according to their specifications) and it appears that at least some common style guides advocate including these kind of dates. Given that we did add a `@dataset` type recently, I'm trying to formulate a policy as to what we are inclined to include or not in the standard styles here.

bastien-roucaries:
Le sam. 18 juil. 2020 à 14:28, moewew <notifications@github.com> a écrit :

> Hmmm, I still have my reservations about this. But they are mostly
> aesthetic: It just seems to be a lot of additional weight for a very, very
> narrow use case. (In particular the required change for the .bst file and
> the data model.)
>
> @plk <https://github.com/plk> What do you think?
>


We have already a patent entry. The problem is the patent style is not
usuable for patent status.

I can improve if needed

moewew:
> The problem is the patent style is not usuable for patent status.

This depends on the style, I'd say. I'm pretty sure APA style would be OK with only one `date`. That's why PLK says this can be left up to the style.

In the standard styles we try to strike a balance between having a usable data model for *basic* use cases and having a clean and understandable data model without too much clutter. The question is whether the four new date fields are so important for referencing patents in a basic/generoc style that it outweighs the burden of adding all the stuff (doing the actual work to add the dates is not the main problem, that is doable and I see you are willing to help, the problem is maintaining the code in the long run and setting a precedent: if we just add anything anyone ever asks about, we'd end up with an unmaintainable mess).

NikolayTach:
Is this issue available on CTAN?

moewew:
@NikolayTach No, this PR hasn't been merged and is therefore not in the release version of `biblatex` (and not on CTAN).

We're still debating whether an extension of the data model on this scale is sensible in the grand scheme of things (see comments above). It has always been a given that the standard data model cannot cover every single use case and that users can add new fields as needed.

PR #928, opened 2019-10-08
added missing translations, few corrections

philip-pirrip:
I am the original translator for the Croatian language (2014 or before). Adding some missing translations and making a few corrections.


moewew:
Thank you very much.

A short run with [`03-localization-keys.tex`](https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/03-localization-keys.tex) shows that apart from the untranslated strings with `FIXME: missing` that still remain (mostly `...total`/`...totals`), the following strings are missing

- `langjapanese`
- `langukrainian`
- `fromjapanese`
- `fromukrainian`

(we don't always add new strings as empty strings with `%FIXME: missing` to all `.lbx` files, sorry, so translators are advised to check their files against [`03-localization-keys.tex`](https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/03-localization-keys.tex)).

Could you push the translation for those four strings as well, please?

If possible the `...total`/`...totals` strings would also be appreciated.

philip-pirrip:
@moewew I did my best, there are no untranslated strings left. I am, however, not sure where and how these ...total/...totals will be used: are they supposed to be in the nominative or genitive case? Croatian also has two forms for plural nouns (1 stih, 2-3-4 stiha, 5 stihova, and again 21 stih). 
Let's just say that this is still better than having untranslated strings and only the corresponding keys printed in bold.

moewew:
Thank you very much.

Unfortunately, the more languages we try to add the more it emerges that the localisation framework we have is too focussed on English, German, French and does not allow to easily capture more intricate features of other languages (for example languages with a more complex number-noun correspondence than just simple 1: singular, everything else: plural).

At some point we will have to revise the localisation system in order to make it easier (or indeed even possible) to translate for a wide variety of languages. But this is a very long-term goal as it would require a huge overhaul of quite a lot of things and it would be crucial that we can get good input to make the framework as flexible and powerful as possible.

PR #936, opened 2019-11-22
Typo 'from' -> 'form'

DavidLuptak:
Hello, I suppose this is a typo in the documentation. Maybe even better wording would be to use `convert`, `map` or `interpret` instead of `form`.


moewew:
Thanks for noticing that.

I guess this should be "will automatically **convert** from roman numerals or arabic letter to integers". So I went with that for https://github.com/plk/biblatex/commit/083bcba47164eb361852be79b3009dfcd08690a9.

Issue #937, opened 2019-11-26
Serbian localization: how to handle variants and settings?


randrej:
I want to contribute a localization for the Serbian language, but first I need
to ask about some things.

Serbian language uses (equally) both Latin and Cyrillic scripts.
This is handled by Babel as two different 'languages': `serbian` is the Latin
variant and `serbianc` is the Cyrillic variant.
If I provide (locally) two `lbx` files, `serbian.lbx` and `serbianc.lbx`, this
works just fine.
Is this the proper way to do it?
The Cyrillic variant differs from the Latin one only in localization strings,
should I inherit it and change the strings?

There is another variable in Serbian language: the dialect or "pronunciation".
The two most used dialects are Ekavian (used in most of Serbia) and Ijekavian
(used in Montenegro, Bosnia and Herzegovina, and parts of Serbia).
They differ in spelling (and pronunciation) for some words containing
"e/je/ije", such as:

*English*: visited, Germany, report  
*Serbian Ekavian*: posećen, Nemačka, izveštaj  
*Serbian Ijekavian*: posjećen, Njemačka, izvještaj

This is handled in `babel-serbian` by the `ijekav` language attribute, so you
may include babel as such:

```tex
usepackage[english, serbian.ijekav]{babel}
% or
\usepackage[english, serbian]{babel}
\languageattribute{serbian}{ijekav}
```

How can I provide such a setting for biblatex?
Is there a default way to provide such settings?

There are also a couple variations of the date format I'd like to provide since
they are used interchangeably and may both used in bibliographies.

I wrote the [`datetime2-serbian`
package](https://gitlab.com/andrejr/datetime2-serbian) with the help of Python,
Jinja2 and my Serbian Cyrillic ↔ Latin transliteration tool
([srtools](https://gitlab.com/andrejr/datetime2-serbian)), and I'll do the same
for biblatex:
write the Latin localization strings (in a YAML file) and then generate the
Cyrillic strings from those.
For `datetime2-serbian` I also generated the LICR/ASCII files from the UTF-8
encoded files using another utility I wrote
([utf8_to_licr](https://gitlab.com/andrejr/utf8_to_licr)).
Is there a way to provide both UTF-8 and ASCII strings in biblatex?

moewew:
`biblatex` currently simply uses `babel` identifiers. So if Latin-script Serbian is `serbian` and Cyrillic is `serbianc` in `babel`-speak, then `serbian.lbx` and `serbianc.lbx` are the way to go.

[...]

PR #940, opened 2019-12-03
Add Serbian localization

randrej:
As announced in #937, here's a Serbian (Cyrillic and Latin) localization for Biblatex.

The files are generated from a common template and YAML string collection by a script.
All of these files can be found at <https://gitlab.com/andrejr/biblatex-serbian>.

Please don't modify these files directly as they are generated - contribute your changes upstream and then send a pull request with the generated file to this repo.

This is done so we could generate both `serbian.lbx` (Latin) and `serbianc.lbx` (Cyrillic) from the same set of localization strings and with the same date format, ordinal format, and other extras.

### DIFF
--- a/tex/latex/biblatex/lbx/serbian.lbx
+++ b/tex/latex/biblatex/lbx/serbian.lbx
@@ -1,0 +1,437 @@
+\ProvidesFile{serbian.lbx}
+[\abx@lbxid]
+
+% Translator's notes:
+%
+% This file is generated from a template and YAML localization string
+% collection by a script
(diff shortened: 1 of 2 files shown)


moewew:
Thank you very much! Merged.

Do you want to be mentioned in the acknowledgment section of the manual? If so, what is the name you would like to see?

randrej:
Sure, why not.
Andrej Radović is my name spelled properly in Serbian, so that will do.

moewew:
Done in https://github.com/plk/biblatex/commit/c98f915b378b3148a4d548894e2dd79ade6bd18c. Thanks again!

Issue #914, opened 2019-08-25
Better timing for \section patch in standard classes


moewew:
MWE adapted from https://tex.stackexchange.com/q/488547/35864
```latex
\documentclass{book}
\usepackage{lipsum}

\usepackage[american]{babel}
\usepackage[utf8]{inputenc}
\usepackage{csquotes}

\makeatletter
\renewcommand{\@chapapp}{Segment}
\makeatother

\usepackage{nameref}  

\usepackage[T1]{fontenc}
\usepackage[
    backend=biber,
    refsegment=subsection+,
]{biblatex}

\defbibheading{subbibliography}{%
  \subsection*{References for
    Segment~\ref{refsegment:\therefsection\therefsegment}:
    \nameref{refsegment:\therefsection\therefsegment}}}

\addbibresource{biblatex-examples.bib}

\usepackage[]{hyperref}
\hypersetup{
    colorlinks=false,
}

\begin{document}
\tableofcontents

\chapter{First, there was light}

\lipsum[0-1]
Cite an author not previously cited~\cite{cicero}.
Note: Alphabetcially, K comes before S.

\section{A Section}

\lipsum[0-1]
Cite an author not previously cited~\cite{cicero}.
Note: Alphabetcially, K comes before S.

\chapter{Then, there was a power outage}

\lipsum[0-1]
And then again cite some authoprs cited previously~\cite{kastenholz}.
At vero eos et accusam et justo duo dolores et ea rebum~\cite{sigfridsson}.

\printbibheading
\bibbysegment[heading=subbibliography]
\end{document}
```

Produces the following bibliography headings

>References for Segment 1: First, there was light
>
>References for Segment 1: First, there was light
>
>References for Segment 2: Then, there was a power outage

but the second heading should read

> References for Segment 1.1: A Section

The problem is the exact timing of the patch that inserts `\newrefsegment` into `\section`-like commands. Better timing is available for `memoir` and from the upcoming version 3.27 also in KOMA Script classes, but the standard styles are trickier. 

Relevant https://github.com/plk/biblatex/pull/887 and https://github.com/plk/biblatex/issues/857

PR #947, opened 2019-12-21
Change timing of section patch for standard classes (#914)

moewew:
See #914. This should make sure that the timing of the patches is better in the standard classes.

### DIFF
--- a/tex/latex/biblatex/biblatex.sty
+++ b/tex/latex/biblatex/biblatex.sty
@@ -10151,9 +10151,10 @@
 %% that we can use for that purpose as well.
 %% This is fairly new, so we need to have a fallback for older versions.
 
-%% The standard classes are tricky. They have no good hooks and just
-%% for sections at least \pretocmd-ing executes the code before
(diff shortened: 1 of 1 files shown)


moewew:
@plk I haven't merged this yet, because I would like to run the test suite first, but I'm currently getting a lot of failures.

I would have expected some failures related to the BibTeX-only change in https://github.com/plk/biblatex/commit/189bc2acb71921fffd5cc9d2ef37b6821b8d4c1c, which should mean that set children only appear in their parent in the bibliography (as is already the case with Biber), but I'm also getting some slightly changed line breaks here and there and even reports of failures on pages that look pretty much the same to me.

plk:
This is normal if you have updated Luatex or formats etc. - kerning changes. Just merge and I'll reconcile those changes in the test suite later. I use a visual PDF diff program to check the actual changes if any are reported.

moewew:
OK, merged the changes, if I broke anything in the tests, give me a shout.

PR #953, opened 2020-01-04
replaced most  occurences; fixes #952

### LINKED ISSUE #952: Brackets in `title` key of `\printbibliography` not correctly handled.
The current implementation of `biblatex` often uses delimited arguments with unknown content without further precautions (e.g., it uses `\foo[#1]` to forward argument `#1` to a macro `\foo` that grabs its argument delimited by `[]`). This can lead to problems when `#1` contains a closing `]`. One example of this issue is, if one provides the following argument to `\printbibliography`:

```latex
\printbibliography[title={Title [Title]}]
```

The braces will be stripped by the key=val-parser (which is the normal behaviour), but `Title [Title]` will later be forwarded as is to `\blx@head@bibliography` as the optional argument without further precautions, so that the argument will be given as

```latex
\blx@head@bibliography[Title [Title]]
```

instead of

```latex
\blx@head@bibliography[{Title [Title]}]
```

Skillmon:
I just searched for occurrences of `[#` (outside of the parameter text of `\def`) and replaced those occurrences with properly braced ones like `[{#2}]` or `[{#2date}]`. I did that without too much code digging, it should work out of the box (and a very small test on the example document from https://tex.stackexchange.com/questions/522902 did indeed work) but I can't guarantee that this won't break things.

This should fix #952 .

### DIFF
--- a/tex/latex/biblatex/biblatex.sty
+++ b/tex/latex/biblatex/biblatex.sty
@@ -3180,7 +3180,7 @@
        {\bibstring{#1s}\ppspace}
        {\def\pno{\bibstring{#1}}%
         \def\ppno{\bibstring{#1s}}}}%
-  \blx@mkpageprefix@i[#2]{#3}}
+  \blx@mkpageprefix@i[{#2}]{#3}}
 
 \long\def\blx@mkpageprefix@i[#1]#2{#1{#2}\endgroup}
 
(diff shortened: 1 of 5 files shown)


moewew:
Manually merged a slightly modified  version of this PR (https://github.com/plk/biblatex/commit/592ca90c68c6e4a5c70e2a5079da19dfaa835e92, I added one more pair of braces in `blx-compat.def` and changed the commit message) in https://github.com/plk/biblatex/commit/35a50374b5545d36c9a2c760c75b6d9a5554d73c.

moewew:
Ooops, forgot to say thank you: Thank you very much for the analysis and the quick fix!

Skillmon:
No problem. I agree that the commit message was suboptimal (it made sense in that second, but will not be of any help in the future). Thanks for your time and effort maintaining this!

PR #958, opened 2020-01-05
Czech and Slovak lbx improvements

DavidLuptak:
Reviewed and completed Czech and Slovak localisations with help of @michal-h21.

Plus some minor fixes in other related files.


moewew:
Thanks. Merged.

@plk This will result in test failures for 03-localization-keys, but they should only involve adding the missing space after "i.e.".

Issue #900, opened 2019-06-23
Compress set references in numeric-comp


moewew:
Compress set references as well

```latex
\documentclass[british]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=numeric-comp, subentry, backend=biber]{biblatex}


\addbibresource{biblatex-examples.bib}


\begin{document}
\cite{set,stdmodel}

\cite{set,sigfridsson,worman}

\cite{herrmann,aksin,yoon}

\cite{herrmann,aksin}

\cite{herrmann,yoon}

\printbibliography
\end{document}
```

See https://tex.stackexchange.com/q/416444/35864

PR #963, opened 2020-02-02
Compress subentry set references in numeric-comp (#900)

moewew:
See #900. Compress subentry set references in `numeric-comp` citation style. Also add a few more configurable delimiters.

This makes the definitions slightly more complex, but I think this is worth it - it makes sense to compress `subentry` citations in a comp style.

Consider
```latex
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=numeric-comp, subentry, backend=biber]{biblatex}

\begin{document}
\cite{set,stdmodel}

\cite{set,sigfridsson,worman}

\cite{herrmann,aksin,yoon}

\cite{herrmann,aksin}

\cite{herrmann,yoon}

\printbibliography
\end{document}
```

---

Technically this is a breaking change, the previous behaviour can be easily restored with the option
```latex
subentrycomp=false,
```
i.e.
```latex
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=numeric-comp, subentry, subentrycomp=false, backend=biber]{biblatex}

\begin{document}
\cite{set,stdmodel}

\cite{set,sigfridsson,worman}

\cite{herrmann,aksin,yoon}

\cite{herrmann,aksin}

\cite{herrmann,yoon}

\printbibliography
\end{document}
```

### DIFF
--- a/doc/latex/biblatex/CHANGES.md
+++ b/doc/latex/biblatex/CHANGES.md
@@ -28,6 +28,17 @@
   `biblatex` no longer falls back to English for unknown languages.
   Warnings will be triggered if undefined language strings or extras
   are used.
+- Added `subentrycomp` option to `numeric-comp` citation style.
+  The option is only relevant with `subentry=true`.
+  With `subentrycomp=true` set citations will be compressed
+  to "1a-c" instead of "1a; 1b; 1c".
+  The option is mainly intended for backwards compatibility,
+  the behaviour of previous `biblatex` versions can be restored
+  with `subentrycomp=false`.
+- Added `\multiciterangedelim`, `\multicitesubentrydelim`,
+  `\multicitesubentryrangedelim`, `\superciterangedelim`,
+  `\supercitesubentrydelim`, and `\supercitesubentryrangedelim` for
+  finer control over (compressed) subentry citations in `numeric-comp`.
 
 # RELEASE NOTES FOR VERSION 3.14
 - biber from version 2.14 has extended, granular XDATA functionality to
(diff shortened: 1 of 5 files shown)


moewew:
I made `subentrycomp=true,` the default because I believe that is what people want from `numeric-comp`, even though that results in changes in the output. People can easily back out of this with `subentrycomp=false,` and I believe that the number of users affected is quite small anyway. But if anyone feels strongly about it I can back out of that and go for `subentrycom=false,` as the default to make sure there is no change in output unless the user specifically requests it.

There will be a test failure for `31-style-numeric-comp`.

josephwright:
Excellent: thanks

plk:
When you merge this, I'll update the test.

moewew:
OK. I'll wait until tomorrow or maybe the day after that to make sure I haven't forgotten anything and then I'll merge if no one objects.

moewew:
Merged. Fingers crossed that I didn't brake anything and that people don't find the names I chose for the ned delimiters too confusing.

If anyone has change suggestions we can still get them in before a new release.

moewew:
I just pushed https://github.com/plk/biblatex/commit/fa6b52e9a384d33b24513d97889ca638eb204ce1 which separates the printing and the compressing bits to allow for easier customisation.

People patching stuff may of course get into trouble here, but I hope this only affects `biblatex-ieee` and I hope to be able to offer a stable solution for that soon: https://github.com/josephwright/biblatex-ieee/issues/51

PR #973, opened 2020-03-15
Add file hooks

moewew:
See https://github.com/josephwright/beamer/issues/581.

Add file hooks so packages/document classes can directly apply changes to arbitrary files before or after they are loaded.

Things are a bit tricky for `.lbx` files, since they can be loaded multiple times. So they have their own interface. I'm not completely happy with that, but it was the best I could come up with. Suggestions for a smoother solution would be highly appreciated.


plk:
When this is commited to DEV, I will merge to the multiscript branch and run the tests as the multiscript branch can load `.lbx` files in a lot more places still due to per-field language switching ...

moewew:
Merged. I don't expect any issues with language loading, but tell me if I broke anything.

Issue #972, opened 2020-03-13
Lithuanian localisation


maras:
Hello.

First of all, thank you for your hard work.

I prepared Lithuanian localisation file, it can be found at: https://github.com/maras/lithuanian.lbx/blob/master/lithuanian.lbx It is based on older unofficial variant which I had to rewrite significantly.
langserbian and fromserbian are commented out because I can't compile file 03-localization-keys.tex with them: 
> Package keyval Error: langserbian undefined. 

> Package keyval Error: fromserbian undefined.

I inserted and commented out langlithuanian.

There is need to write the word "language" next to the name of a language that is why next to the language name is word "kalba" or abbreviation "k." in the language strings.
There is declension in Lithuanian, so in some cases there is need to use genitivus (translated from). 
This is especially important working with month names, so I created local forms of month in genitivus (they are commented out now).

## Answers to the localisation questions.

> What kind of dash do you use in number ranges like "35-50"?

 `\textendash` There are no any spaces around the dash.

> What kind of dash do you use in date ranges like "1st January-3rd February 2009"?

`\textendash` and there *are* spaces around the dash.

> In lists and enumerations, do you add a comma in addition to connectors like "and"? 

No, there is no comma before last "and".

>    How do you format short ordinals? Are ordinals gender/number sensitive? 

All ordinals are gender sensitive and number insensitive:
1-as 2-as 3-as... (masc.), 1-a 2-a (fem.)

> What's the gender of the word "edition"?

Masculine.

> What's the gender of the word "series" (of a journal)?

Feminine.

> What's the long date format?

[year] m. [genitivus of month name] [day] d.
2020 m. kovo 4 d.

> What happens if the day is missing from the date?

[year] m. [nominativus of month name]
2020 m. kovas

> What happens if both day and month are missing?

[year] m.
2020 m.

> Does this day format use leading zeros?

No.

> Please be specific about the spaces to be used (non-breakable, thin spaces).

I don't know about specific requirements concerning spaces in the long dates.

> What happens for date ranges with compact (1.--6. February 2010, 1. February--6. March 2010) and non-compact ranges (1. February 2010--6. February 2010)?

- Compact (they are quite rare): 2020 m. kovo 13--14 d. (endash between days without spaces, month – genitive) 
  But when there are no days: 
  2020 m. sausis -- kovas or
  1920 m. sausis -- 1922 m. kovas
  (endash with spaces between month, month – nominative)
  The same when only year: 1920 m. -- 1922 m.
[...]

PR #780, opened 2018-08-15
Adds Hungarian localization to biblatex 

priestoferis:
and updates existing localizations with langhungarian and fromhungarian
(some of them just placeholder comments, others translated and confirmed
by native speakers)

created by:
- moewew
- Márton Marczell
- Bence Ferdinandy

### squashed commits below ###

first few translations.

fix encoding. todo: convert to ascii and tex accents when done

.

Functions done

actions

.

Implement Hungarian word order for pages

Add dots for page numbers

Add more dots for hungarian page numbers

.

Translated (almost) all strings

backrefpages

Change volume field order

Ordinal volumes and chapters

Rewok page numbering

- Redefine fewer low-level commands and focus on higher-level
  definitions instead. This makes the Hungarian changes more persistent
  but also more disrupting and possibly descructive for custom styles.
- Document possibly discrupting behaviour and add warnings in the lbx.

Implement Hungarian dates and bibextras

Fix undefined \lbx@hu@dayrange with fallback

Some improvements

Fix capitalisation for empty origlanguage

Polish magyar.lbx

Unpack definition of \BiblatexHungarianWarningOff

improvements

changed accented character to LaTeX comments

change \H o to \H{o} and \H u to \H{u}

switch ü-s to \"u-s

fixed ú

A few fixes

- updated branch from dev
- add langhungarian and fromhungarian to the available languages
- added some translations I could confirm with native speakers

removed encoding from header

filled out ukrainian

galician, russian and spanish added

adding hun to brazilian

PR #977, opened 2020-03-21
Add Lithuanian localisation (#972)

moewew:
See #972.

Includes small tweaks to Hungarian localisation for uniformity (ping back to #717, #780).


moewew:
@plk I just realised that `lithuanian.lbx` uses UTF-8 unlike most other Latin-script-based `.lbx` files. Should we re-code non-US-ASCII characters into macros or should we stick with UTF-8 and add a note about that in the docs and a warning in the `.lbx` itself (as we do for Cyrillic-based `.lbx` files).

plk:
Let's stick to UTF-8 - the more we encourage this, the better.

moewew:
OK. Added a note about this in the docs

Issue #985, opened 2020-04-06
Check @set entries with BibTeX


moewew:
`@set` entries in non-numeric styles are always tricky, but
```latex
\documentclass[british]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=authoryear, backend=bibtex]{biblatex}


\addbibresource{biblatex-examples.bib}


\begin{document}
\autocite{set,stdmodel}
\printbibliography
\end{document}
```
and
```latex
\documentclass[british]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=alphabetic, backend=bibtex]{biblatex}


\addbibresource{biblatex-examples.bib}


\begin{document}
\autocite{set,stdmodel}
\printbibliography
\end{document}
```
need looking at.

Can we get them to sort by the first entry in BibTeX? (Like Biber does.)

moewew:
@plk https://github.com/plk/biblatex/commit/5e3c8632f5618cbee9ee30e40e038b3a389bb51e should hopefully fix the `extradate` test failure you mailed me about.

But really sets are severely broken for anything other than `style=numeric, sorting=none,` in BibTeX: BibTeX can't sort them correctly and can't generate usable labels for anything else. I noticed that `biblatex` v1.0 can deal better with BibTeX entry sets than v3.14, so I'll see if we can take something from there.

---

Ping back to #921

PR #986, opened 2020-04-10
Improve @set sorting for BibTeX (#985)

moewew:
See #985.

Implement a two-pass structure with internal copies of `@set` entries to be able to sort `@set`s properly.

This now requires additional BibTeX runs so that the sequence becomes LaTeX, BibTeX, LaTeX, BibTeX, LaTeX, LaTeX if `@set` s are involved. The tests may have to be adjusted accordingly so that they don't stop in the middle of a half-compiled document.

### DIFF
--- a/bibtex/bib/biblatex/biblatex-examples.bib
+++ b/bibtex/bib/biblatex/biblatex-examples.bib
@@ -25,19 +25,13 @@
 
 @set{set,
   entryset     = {herrmann,aksin,yoon},
-  annotation   = {A \texttt{set} with three members. The \texttt{crossref} field
-                  in the \texttt{@set} entry and the \texttt{entryset} field in
-                  each set member entry is needed only when using BibTeX as the
-                  backend},
+  annotation   = {A \texttt{set} with three members.},
 }
 
 @set{stdmodel,
   entryset     = {glashow,weinberg,salam},
   annotation   = {A \texttt{set} with three members discussing the standard
-                  model of particle physics. The \texttt{crossref} field
-                  in the \texttt{@set} entry and the \texttt{entryset} field in
-                  each set member entry is needed only when using BibTeX as the
-                  backend},
+                  model of particle physics.},
 }
 
 @article{aksin,
--- a/bibtex/bst/biblatex/biblatex.bst
+++ b/bibtex/bst/biblatex/biblatex.bst
@@ -16,6 +16,7 @@
 
 ENTRY {
     % special fields
+    fakeset
     entryset
     entrysubtype
     execute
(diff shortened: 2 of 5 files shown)


plk:
Do you know which tests need additional runs?

moewew:
All BibTeX tests that involve `@set`s. Since BibTeX runs are *really* cheap we can probably afford running all tests with the longer sequence, but I can check which tests are really affected if you think that is worth it.

moewew:
If I didn't miss anything the following BibTeX tests will need more runs.

Tests with `\nocite{*}`
- 01
- 02
- 03
- 16
- 17
- 19
- 20
- 21
- 40
- 50
- 60
- 92

Tests citing `set` or `stdmodel`
- 30
- 31
- 32

plk:
Yes, fair enough, I'll modify all tests.

plk:
I modified the tests and ran the bibtex suite and everything passed against the current reference docs - was the sorting order supposed to change in e.g. the `50-*` test files?

moewew:
Mhh, yes. The BibTeX output should look like this now

[50-style-authoryear.pdf](https://github.com/plk/biblatex/files/4462066/50-style-authoryear.pdf)

Specifically, the `@set` entries should not sort first any more in the bibliography. Compare with the current version https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/50-style-authoryear-bibtex.pdf, where the first entry is the "Herrmann, Wolfgang A" set and the second "Glashow, Sheldon".

plk:
I think I didn't update my local install with the merge - running again ...

plk:
Looks good - I did notice that in `02-*` the labels looks a little odd:

```
blx@fakeset-stdmodel
```

moewew:
Ah, yes. That is a side effect of the "hidden copy" of the `@set` entry. Do you think this is something I should try to fix (I thought about that before: it might be possibly to overwrite the entrykey again at some point, but I'm not sure if that will not cause other issues)?

plk:
I think it's fine - I'll push 3.15 again - thanks for doing that - it makes the test files look odd.

moewew:
I just pushed https://github.com/plk/biblatex/commit/b51c747397893170906e62f423e7ace9085faf67, which should take care of some weird edge cases and should also get the correct entry keys in the test `02-`. If you could run the tests again and tell me if I broke something that would be grand.

moewew:
Oh, hang on `alphabetic` labels are gone again. I'll look into it, push a fix and ask you to run the tests after that.

moewew:
Hopefully https://github.com/plk/biblatex/commit/300c0f1ad3b184abd2c864f9c0f66c132864c1cc fixes that. Pushed, so you can run the tests.

plk:
Looks better - test files updated and new dev/multiscript packages pushed.

PR #991, opened 2020-04-13
Abstract key-val interface

moewew:
This should make it easier to switch to a different keyval package than `keyval` (something category stable that doesn't break with Turkish) in the future.

We use `\provide...` (or equivalent constructs) to overwrite existing definitions of `\blx@kv@...` which means users can define their own replacements before loading `biblatex`.

@Skillmon may for example want to use `expkv` with `biblatex`:
```latex
\documentclass[turkish]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}


\usepackage{expkv}
\makeatletter
% define a new key (possibly with a default value)
% {<family>}{<key>}[<default>]{<code>}
\newcommand*{\blx@kv@defkey}[2]{%
  \@ifnextchar[%
    {\blx@kv@defkey@default{#1}{#2}}
    {\blx@kv@defkey@nodefault{#1}{#2}}}

\def\blx@kv@defkey@nodefault{\ekvdef}
\def\blx@kv@defkey@default#1#2[#3]#4{%
  \csdef{blx@opthelper@#1@#2}##1{#4}%
  \ekvdef{#1}{#2}{\csuse{blx@opthelper@#1@#2}{##1}}%
  \ekvdefNoVal{#1}{#2}{\csuse{blx@opthelper@#1@#2}{#3}}}

% globally define new key (should only be the last resort)
% {<family>}{<key>}[<default>]{<code>}
\newcommand*{\blx@kv@gdefkey}[2]{%
  \@ifnextchar[%
    {\blx@kv@gdefkey@default{#1}{#2}}
    {\blx@kv@gdefkey@nodefault{#1}{#2}}}

\def\blx@kv@gdefkey@nodefault{\global\ekvdef}
\def\blx@kv@gdefkey@default#1#2[#3]#4{%
  \csgdef{blx@opthelper@#1@#2}##1{#4}%
  \global\ekvdef{#1}{#2}{\csuse{blx@opthelper@#1@#2}{##1}}%
  \global\ekvdefNoVal{#1}{#2}{\csuse{blx@opthelper@#1@#2}{#3}}}

% test if key is undefined
% {<family>}{<key>}{<true>}{<false>}
\newcommand*{\blx@kv@ifkeyundef}[2]{%
  \ekvifdefined{#1}{#2}
    {\@secondoftwo}
    {\@firstoftwo}}

% undefine key
% {<family>}{<key>}
\newcommand*{\blx@kv@undef}[2]{\csundef{\ekv@name{#1}{#2}}}
 
% {<family>}{<key-vals>}
\newcommand*{\blx@kv@setkeys}{\ekvset}
\makeatother

\begin{document}
\cite{sigfridsson}
\printbibliography
\end{document}
```

### DIFF
--- a/doc/latex/biblatex/CHANGES.md
+++ b/doc/latex/biblatex/CHANGES.md
@@ -84,6 +84,16 @@
   and friends). Those wrappers make no sense, since the indexed
   names are not printed 'together' in any useful sense of the word
   and were never working anyway.
+- **CRITICAL CHANGE** Generalised some keyval macros.
+  `biblatex`-related uses of `\define@key` should ideally replaced with
+  `\blx@kv@defkey`.
+  Style authors are advised not to rely too much on the internal
+  implementation of certain keyval interfaces.
+  Users who want to experiment with using a different underlying
+  keyval parser need only provide replacements of the `\blx@kv@...`
+  macros (which are defined via `\provide...` so that they will
+  not overwrite existing definitions; in particular users can define
+  those replacements before loading `biblatex`).
 
 # RELEASE NOTES FOR VERSION 3.14
 - biber from version 2.14 has extended, granular XDATA functionality to
--- a/tex/latex/biblatex/biblatex.def
+++ b/tex/latex/biblatex/biblatex.def
@@ -18,9 +18,9 @@
 % \namepart<namepart>i - e.g. \namepartfamilyi
 \def\do#1{%
   \csdef{mkbibname#1}##1{##1}%
-  \define@key{blx@opt@namepart}{#1}{\csdef{namepart#1}{##1}}%
-  \define@key{blx@opt@namepart}{#1i}{\csdef{namepart#1i}{##1}}%
-  \define@key{blx@opt@namepart}{#1un}{\csdef{namepart#1un}{##1}}}
+  \blx@kv@defkey{blx@opt@namepart}{#1}{\csdef{namepart#1}{##1}}%
+  \blx@kv@defkey{blx@opt@namepart}{#1i}{\csdef{namepart#1i}{##1}}%
+  \blx@kv@defkey{blx@opt@namepart}{#1un}{\csdef{namepart#1un}{##1}}}
 \expandafter\docsvlist\expandafter{\blx@datamodel@constant@nameparts}
 
 % legacy aliases
--- a/tex/latex/biblatex/biblatex.sty
+++ b/tex/latex/biblatex/biblatex.sty
@@ -53,6 +53,39 @@
    \endinput}
 
 
+% generalised keyval interface
+% define a new key (possibly with a default value)
+% {<family>}{<key>}[<default>]{<code>}
+\providerobustcmd*{\blx@kv@defkey}{\define@key}
+
+% globally define new key (should only be the last resort)
+% {<family>}{<key>}[<default>]{<code>}
+\providerobustcmd*{\blx@kv@gdefkey}[2]{%
+  \@ifnextchar[%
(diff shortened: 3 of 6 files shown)


plk:
@moewew - running tests after this (and the previous index wrappers commit), I get some differences in 20,21,22 and 31 test files. Attached is screenshot diff of the first differences in 20 - some indices are missing?

Screenshot 2020-04-13 at 1 05 pm

moewew:
@plk Oh sorry. Apparently I removed a line I shouldn't have removed in https://github.com/plk/biblatex/commit/95c86f3b52d8f0a2bd3fa3e78a2cd98a55d395ec, should be fixed in https://github.com/plk/biblatex/commit/7bbc2155cce69ba43cef1fdd1d845a38a810d328. Fingers crossed. Thanks for catching that!

Skillmon:
@moewew Very good example:) Just want to point out that the macro `\newcommand*{\blx@kv@undef}[2]{\csundef{\ekv@name{#1}{#2}}}` should've been `\newcommand*{\blx@kv@undef}[2]{\csundef{\ekv@name{#1}{#2}}\csundef{\ekv@name{#1}{#2}N}}` to also clear the default if it was set. Do you intend to provide options to use different key=val implementations without it requiring advanced knowledge from the user (else I don't see much advantage from this generalisation)?

moewew:
Fixed, thank you.

Eventually I was hoping to give users a way to switch more easily, yes. But I haven't really tried to figure out what how we could best do that and it would probably need additional testing of the alternatives. For now this PR mainly makes sure that we understand what we need from a key-val interface and gives us a unified way to use the required features.

Skillmon:
@moewew I'd provide single files for each supported key=val interface (so that new ones can be added easily by pull requests) and a loader-package which gets a single option stating the used key=val package. Something like the following:

```tex
\ProvidesPackage{blxkeyval}
\DeclareOption*{\let\blxkeyval@pkg\CurrentOption}
\ProcessOptions\relax

\IfFileExists{blxkeyval.\blxkeyval@pkg.tex}
  {\input{blxkeyval.\blxkeyval@pkg.tex}}
  {\PackageError{blxkeyval}{Unknown key=val package `\blxkeyval@pkg'}{}}

\endinput
```

Your code above could then be the contents of `blxkeyval.expkv.tex`.

plk:
@moewew - I wonder if you could have a look at biblatex.sty in the multiscript branch - I merged the keyval changes but there are a couple of places where I def raw `KV@` in some deeply nested defs and I'm not sure how to integrate this with the new macros - it is likely more obvious to you if the keyval code is still in your head for a while.

moewew:
I hope https://github.com/plk/biblatex/pull/992 is enough, but I may well have missed something.

moewew:
I didn't expect this change to have impact on contributed styles, but just grepped for `\define@key` in my directory of `biblatex` styles and noticed that `biblatex-philosophy` uses it to add custom options to `\printbibliography`. Maybe we need a real interface for that?

Anyway for now, style authors will want to replace `\define@key` with `\blx@kv@defkey` and `\setkeys` with `\blx@kv@setkeys`. If they want backwards compatibility with older versions of `biblatex` (probably a good idea) they'll probably also want to add
```latex
\providerobustcmd*{\blx@kv@defkey}{\define@key}
\providerobustcmd*{\blx@kv@setkeys}{\setkeys}
```

I think the developer of `biblatex-philosophy` is ivalb on GitHub but just to be sure I got the right person, I'll send an email and avoid `@` mentioning him here.

PR #994, opened 2020-04-14
Fix Swedish translation of `nodate`.

adriangoransson:
Below are the links to some university style guides all suggesting this way of writing.

I'm unsure if the `å` is correct (I haven't tested it) since it's written `\aa ` including the trailing space before `\adddot`. This seems to be the how it's written elsewhere though.

Umeå universitet, Harvard - hänvisningar i text
https://www.umu.se/bibliotek/soka-skriva-studera/skriva-referenser/harvard-hanvisningar-i-text/

Karolinska Institutet, Referensguide för APA
https://tools.kib.ki.se/referensguide/apa/#webb-webbsidor

Jönköpings universitet, APA – Referenshantering
https://guides.library.ju.se/c.php?g=668119&p=4842879

Röda korsets högskola, Guide till referenshanteringenligt APA-systemet
https://www.sh.se/download/18.7d9f1b25167c61c5af38f695/1551022923629/APA_referenser_2015.pdf

### DIFF
--- a/tex/latex/biblatex/lbx/swedish.lbx
+++ b/tex/latex/biblatex/lbx/swedish.lbx
@@ -336,7 +336,7 @@
   origpubin        = {{ursprungligen publicerad}{ursprungligen publ\adddot}},
   astitle          = {{som}{som}},
   bypublisher      = {{av}{av}},
-  nodate           = {{u\adddot d\adddot}{n\adddot d\adddot}},%FIXME
+  nodate           = {{utan \aa rtal}{u\adddot \aa \adddot}},
   page             = {{sidan}{s\adddot}},
   pages            = {{sidorna}{s\adddot}},
   column           = {{spalt}{sp\adddot}},
--- a/tex/latex/biblatex/lbx/swedish.lbx
+++ b/tex/latex/biblatex/lbx/swedish.lbx
@@ -336,7 +336,7 @@
   origpubin        = {{ursprungligen publicerad}{ursprungligen publ\adddot}},
   astitle          = {{som}{som}},
   bypublisher      = {{av}{av}},
-  nodate           = {{utan \aa rtal}{u\adddot \aa \adddot}},
+  nodate           = {{u\adddot\,\aa \adddot}{u\adddot\,\aa \adddot}},
   page             = {{sidan}{s\adddot}},
   pages            = {{sidorna}{s\adddot}},
   column           = {{spalt}{sp\adddot}},


moewew:
Thank you very much.

A few other Swedish translations seem to be missing as well, if you want to contribute, you can run https://gist.github.com/moewew/f11feb2c822339cc0b0b4f292ece13df (a modified version of https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/03-localization-keys.tex that works with the current release version of `biblatex` v3.14) and see what can be added.

@perstar You contributed to the Swedish translations in the past, if you could spare a moment to look over these changes (and if you can spare some more moments over https://gist.github.com/moewew/f11feb2c822339cc0b0b4f292ece13df as well) that would be very much appreciated.

moewew:
Oh, just to double check.
https://github.com/plk/biblatex/blob/a56f8ab182fa2aaaf53b1af1e1c095601542c853/tex/latex/biblatex/lbx/swedish.lbx#L339
will produce

> u.å.

without any space between the dot after the u and the å. Is that intentional?
https://github.com/plk/biblatex/blob/a56f8ab182fa2aaaf53b1af1e1c095601542c853/tex/latex/biblatex/lbx/swedish.lbx#L309
for example has a thin space between "m." and "fl.".

adriangoransson:
I'm doing the translations as I find them, but I'll see if I can find anything obvious! :)

Checking the above sources and when comparing with "n.d." I'm pretty confident that there should be no space.

`andothers` is most commonly written "et al." from what I gather even in Swedish. It's only the last source in my original commit that mentions "m.fl." and that is without the space. The [swedish wikipedia article about et al](https://sv.wikipedia.org/wiki/Et_al.) agrees.

I assume that the translations are long/short format? In that case "m.fl." should be "med flera".

perstar:
Thanks for adding missing strings!

I agree that "u.å." is certainly the abbreviation used in Swedish.

>> andothers        = {{m.\,fl\adddot}{m.\,fl\adddot}}, 
> I assume that the translations are long/short format? In that case "m.fl." should be "med flera".

That's what the words *mean*, but that doesn't mean you would even want to have them written out in a bibliography. Where that line goes is not obvious for all entries, but it was certainly intentional by me to have the same string in long/short format. (Just as for example english.lbx has the same andother strings for short and long, for example, and never "et alias".)

Similarly (but maybe not 100% sure like for "m.fl.") I don't think "u.å." even should be written out as "utan år" (or "utan årtal") in a bibliography. Even though it's not clear I think the references you cite not intend that you would ever write it out. They just tell you what the abbreviation means. 

> Checking the above sources ... I'm pretty confident that there should be no space

I can't agree with that reasoning. How you actually typeset Swedish abbreviations is not something they talk about at all, but that is something you go to (Swedish) typographical references for. And it's not something done differently abbreviation by abbreviation so that "u.å." and "m.fl." are handled differently! Writing plain texts like what I'm doing now is something very different from producing typeset text, which is why I have no problem writing "m.fl." here even though I would use other quotation marks and some thin space in a "typographical text".

As for "m.fl." vs "et al." see the comment in swedish.lbx about that!

moewew:
@perstar Thank you very much!

@adriangoransson There isn't a lot to add to @perstar's comments from my side. He made the two important points that need to be made here

1. Some strings are almost exclusively used in their short form. In that case it is perfectly fine that long and short form coincide (and are both the short version). One example would be "et al." in English, which I have never seen as "et alii"/"et aliae"/"et alia" in any bibliography (note how the short for "et al." nicely avoids the gender issue here, that can be incredibly useful for some languages). It's the same with German "u.a.", it's not spelled out as "und andere" in the long form either.

    It is most definitely not completely trivial to decide which strings should behave normally and which strings are always short. But `andothers` and `nodate` are certainly good candidates for the *always short* case. To some degree this might be a matter of taste. But if we can find no source that uses the long form "med flere" (except when it explains what "m.fl." means), we're probably on the right track with always using the short form.

2. It's not clear whether the bibliographical recommendations we can find on various (often university) websites claim authority over finer typographical points as well. While it is certainly possible to use more "advanced typographic" features such as thin spaces, non-breaking spaces and curly quotation marks in HTML and MS Word-generated PDF documents, many people don't, so the absence of those features may not necessarily of prescriptive nature (as in "don't use these features"), but may well be the result of the author being oblivious of those finer points.

    In German for example it is quite common to write "u.a." or "z.B." on websites, but in printed texts you'd want to see something like `u.\,a.` or `z.\,B.` and the Duden seems to accept only that (I thought I read a passage that it is OK-ish not to use a thin space if you can't get one, but I can't find a reference for that now, so take that as [citation needed] and probably incorrect).

adriangoransson:
Thank you for your comments @perstar, I'm clearly out of my depth here.

I'll submit a corrected version with short/short strings. How about the space?

> ... it's not something done differently abbreviation by abbreviation so that "u.å." and "m.fl." are handled differently!

To clarify: You're saying that abbreviations should always be treated the same with regards to their spacing? In other words: "u.`thin space`å." and  "m.`thin space`fl." etc?

In that case I will fix that as well.

moewew:
Maybe the date abbreviations needs to be reviewed as well w.r.t. spaces between dots

https://github.com/plk/biblatex/blob/a56f8ab182fa2aaaf53b1af1e1c095601542c853/tex/latex/biblatex/lbx/swedish.lbx#L501-L504

adriangoransson:
Sorry for the late response. I don't have neither knowledge nor time to make an assessment regarding the spacing and punctuation of other strings. I would suggest creating a new issue to go over the current translations and make them consistent. Also to fix short/short formatting where applicable.

`nodate` feels more urgent to me as it's currently not even translated (in the short form).

perstar:
> In other words: "u.thin spaceå." and "m.thin spacefl." etc?

Yes, at least "u.å." should be treated similar to how for example "m.fl." already is treated, that is have a thin space.

moewew:
Merged. Thank you very much, @adriangoransson and @perstar.

PR #998, opened 2020-04-15
`nodate` now conforming to the "Règles typo. en usage à l'IN".

dbitouze:


moewew:
Note that removing the leading `{}` in the short form will lead to the "s" being capitalised as normal. Sometimes this may not be desired (so much so that in some languages this is explicitly suppressed), see
```latex
\documentclass[french]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\DeclareDelimFormat[bib]{nameyeardelim}{\addperiod\space}

@book{appleby,
  author  = {Humphrey Appleby},
  title   = {On the Importance of the Civil Service},
}
\end{filecontents}
\addbibresource{\jobname.bib}


\begin{document}
\autocite{appleby}
\printbibliography
\end{document}
```

dbitouze:
@moewew OK. I don't know how to proceed: submit a new PR?

moewew:
**If** you want to add the `{}` you can just push another commit to this branch, normally the pull request should then update automatically.

Note that I did not want to say that you need to add the `{}`, I'll have to leave that up to you, because I have no idea about French typographic conventions.

dbitouze:
> Note that I did not want to say that you need to add the `{}`, I'll have to leave that up to you, because I have no idea about French typographic conventions. I just wanted to point out the difference.

You did well: the initial of the short form shouldn't be uppercase.

moewew:
Thanks for checking that and for the new commit.

Merged. Thank you very much.

dbitouze:
You're welcome! :smile:

Issue #997, opened 2020-04-15
Spaces in abbreviations in german.lbx


moewew:
Because of the discussion in https://github.com/plk/biblatex/pull/994 I checked out the situation with respect to spaces and dots in `german.lbx` and found that many strings use `.\,`

https://github.com/plk/biblatex/blob/200e3cf74e375d3f4082dc36cbbd0f878c921847/tex/latex/biblatex/lbx/german.lbx#L349

https://github.com/plk/biblatex/blob/200e3cf74e375d3f4082dc36cbbd0f878c921847/tex/latex/biblatex/lbx/german.lbx#L359-L360

but `nodate` has

https://github.com/plk/biblatex/blob/200e3cf74e375d3f4082dc36cbbd0f878c921847/tex/latex/biblatex/lbx/german.lbx#L374

and the other date-related strings have

https://github.com/plk/biblatex/blob/200e3cf74e375d3f4082dc36cbbd0f878c921847/tex/latex/biblatex/lbx/german.lbx#L543-L546

I haven't had a look at my printed Duden yet, but a little Googling suggests that `.\,` is what most people would recommend.

@jspitz Do you have any insight here?

---

Then there is 

https://github.com/plk/biblatex/blob/200e3cf74e375d3f4082dc36cbbd0f878c921847/tex/latex/biblatex/lbx/german.lbx#L424

but

https://github.com/plk/biblatex/blob/200e3cf74e375d3f4082dc36cbbd0f878c921847/tex/latex/biblatex/lbx/german.lbx#L442

and

https://github.com/plk/biblatex/blob/200e3cf74e375d3f4082dc36cbbd0f878c921847/tex/latex/biblatex/lbx/german.lbx#L443-L444

jspitz:
I'd say it depends whether it's an established abbreviation which most people would write with no space. These I would set with thinspace (this includes `o.\,D.` above). I would use a normal (non-breaking) space with `s.~auch` and `s.~Anm.` But I am not a typographer and I understand this is more a matter of taste (I've consulted "Detailtypographie" by Forssman/de Jong, but they are not very clear on this either).

moewew:
@jspitz Thanks for the input.


That sounds like a good approach. Thin spaces make sense in short multi-letter abbreviations, `s.~auch` also looks sensible. For longer abbreviations one could also consider `\adddotspace` instead of `\addnbspace`/`~`, which uses `abbrvpenalty` to discourage but not completely rule out line breaks.

[...]

PR #999, opened 2020-04-16
Rework abbreviations in German.lbx (#997)

moewew:
As discussed in #997
* common multi-letter abbreviations use `\,`
* "s." (siehe) uses `~`
* other abbreviations use `\adddotspace`


jspitz:
Looks good to me.

moewew:
Thanks for the feedback!

Merging because no one else complained.

Issue #996, opened 2020-04-14
Why not add an issuetitleaddon field ?


jeromevillebrun:
Hi, 
For a bibliographic style I am writing, I would need a field that go just after the `issuetitle` of entries which use the `issue` macro. There is already a `titleaddon` field, so why wouldn't there be a `issuetitleaddon` ? It would be very usefull for many french bibliography styles (and maybe other languages). 

In french, a typical example of what would go in this field would be an inscription like "numéro spécial" or "dossier thématique" (I don't know the translation in english, but it doesn't matter since `issuetitleaddon` would be a litteral field, exactly as `titleaddon` is).

If prefered, I can try to do a Pull Request, but for the moment, I have no experience in creating new fields (and I also don't where the documentation is).

moewew:
You can 'easily' add new fields yourself with a new datamodel file. See https://tex.stackexchange.com/q/163303/35864.

To get a `issuetitleaddon` you could say
```latex
\documentclass[british]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\begin{filecontents}{issuetitleaddon.dbx}
\ProvidesFile{issuetitleaddon.dbx}[2020/04/14 issuetitleaddon]

\DeclareDatamodelFields[type=field,datatype=literal]{issuetitleaddon}
\DeclareDatamodelEntryfields[article,periodical]{issuetitleaddon}
\end{filecontents}

\usepackage[style=authoryear, backend=biber, datamodel=issuetitleaddon]{biblatex}

\renewbibmacro*{issue}{%
  \ifboolexpr{
    test {\iffieldundef{issuetitle}}
    and
    test {\iffieldundef{issuesubtitle}}
  }
    {}
    {\printtext[issuetitle]{%
       \printfield[titlecase]{issuetitle}%
       \setunit{\subtitlepunct}%
       \printfield[titlecase]{issuesubtitle}}}%
  \printfield{issuetitleaddon}}

\begin{filecontents}{\jobname.bib}
@article{appleby,
  author          = {Humphrey Appleby},
  title           = {On the Importance of the Civil Service},
  date            = {1980},
  journal         = {Journal of the Civil Service},
  volume          = {234},
  issuetitle      = {Importance Issue},
  issuetitleaddon = {Special Issue 123},
}
\end{filecontents}
\addbibresource{\jobname.bib}
\addbibresource{biblatex-examples.bib}


\begin{document}
[...]

Issue #981, opened 2020-03-31
Unpublished entries do not print eprint information


eddieschoute:
I'm trying to cite an article that has only been prepublished on ArXiv. The suggestions in https://tex.stackexchange.com/questions/456340/unpublished-arxiv-entries-with-biblatex-biber or https://tex.stackexchange.com/questions/49757/what-should-an-entry-for-arxiv-entries-look-like-for-biblatex to use the `@online` type technically work but print the title in italics instead of in quotes, which is how articles are printed. Moreover, the `unpublished` type with `pubstate=prepublished` seems to apply better here.
An example entry
```
@Unpublished{Author2019,
  author      = {Author},
  title       = {My title},
  date        = {2019-12-12},
  eprint      = {1912.9999v3},
  eprintclass = {quant-ph},
  eprinttype  = {arXiv},
  pubstate    = {prepublished},
}
```
prints
> [6] Author. “Mytitle”. Dec.12,2019. Pre-published.

Note that there is not preprint information. Whereas if I change it to an article, then it outputs 
> [6] Author.“Mytitle”. In:(Dec.12,2019). arXiv:1912.9999v3[quant-ph]. Pre-published.

as expected (note the empty journal field causes an empty "In:" to be printed).

I'm also open to other suggestions.

moewew:
It won't surprise you that my advice is to use `@online` for articles on the arXiv, after all that is what `biblatex-examples.bib` does: https://github.com/plk/biblatex/blob/2ddd2836927d08ac600c8fa3af1897417548ac05/bibtex/bib/biblatex/biblatex-examples.bib#L1392-L1405

If you don't like the way the title is formatted, you can easily adjust that with
```latex
\DeclareFieldFormat[online]{title}{\mkbibquote{#1\isdot}}
```

----

I'm not completely opposed to adding `eprint` information to `@unpublished` since the trio `doi+eprint+url` is so widespread nowadays that it should probably apply to all entry types alike. Indeed only, `@online` and `@unpublished` are picky about DOIs and `eprint`s and don't use the bibmacro `doi+eprint+url`

https://github.com/plk/biblatex/blob/0d37281977171eaf945850ef3363fa94e1d44c9a/tex/latex/biblatex/bbx/standard.bbx#L467-L506

https://github.com/plk/biblatex/blob/0d37281977171eaf945850ef3363fa94e1d44c9a/tex/latex/biblatex/bbx/standard.bbx#L701-L734

[...]

Issue #847, opened 2018-12-15
how to enter journal articles deviating from the traditional pages format


mkimki:
I didn't find an answer on this elsewhere, so I hope opening this issue is ok.

There are modern electronic-only journals which deviate from the traditional pages format. Instead, the articles within the same volume get some identifier.
An example is the [Electronic Journal of Combinatorics](http://www.combinatorics.org/). Let's look at this [article](http://www.combinatorics.org/ojs/index.php/eljc/article/view/v19i3p18/pdf). It is given as Volume 19, Issue 3 (2012), Paper #P18. In mathscinet it is given as Electron. J. Combin. 19 (2012), no. 3, Paper 18, 12 pp. 

How should we add such references with biblatex?
Type should be article, I guess. The closest thing I can think of is pages = "\#P18", pagetotal = "12 pp.". However, "\#P18" might couse formatting problems at some places since it is not a page number or a page number range. And the field "pagetotal" is not supported by the type article, at the moment.

moewew:
There is a field for `@article` entries called `eid` that can be used for the 'pseudo-pagination' of electronic journals. There is indeed no `pagetotal` for `@article`s, if the total number of pages is important, you could use the `pages` field.

```
\documentclass[british]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=authoryear, backend=biber]{biblatex}

\usepackage{filecontents}
\begin{filecontents}{\jobname.bib}
@article{betsumiya,
  author  = {Koichi Betsumiya and Masaaki Harada and Akihiro Munemasa},
  title   = {A complete classification of doubly even self-dual codes of length 40},
  journal = {The Electronic Journal of Combinatorics},
  volume  = {19},
  number  = {3},
  eid     = {P18},
  date    = {2012},
  pages   = {1-12},
}
\end{filecontents}

\addbibresource{\jobname.bib}
\addbibresource{biblatex-examples.bib}


\begin{document}
\cite{sigfridsson,betsumiya}
\printbibliography
\end{document}
```

mkimki:
Thank you for this detailed answer! I was not aware of the eid entry.

In the presence of the pagetotal field, using the pages field for the number of pages feels a bit like cheating. Is there a particular reason why there is no pagetotal for @article? Would it make sense to add it?

moewew:
[...]

PR #1000, opened 2020-04-17
Changes to standard drivers/bibmacros

moewew:
A selection of changes to the standard drivers and bibmacros (combined here for easier discussion). I hope those changes are useful and don't pose issues to contributed styles.

Feel free to comment on any of the points mentioned here.

## Unify DOI, eprint and URL printing (#981)

Until now `@online` and `@unpublished` were the only entry types that did not use the bibmacro `doi+eprint+url` to print DOI, eprint and URL. This made sense for `@online` since it did not obey the global `url` option (URLs in `@online` entries are always printed even with `url=false`) and `@unpublished` only allowed for URLs. Since DOIs and eprint are more widely available nowadays, it seems useful to allow these kinds of links for all entry types.

Since the options `doi`, `eprint` and `url` are available in per-type scope now, it is possible to make even `@online` use the `doi+eprint+url` macro, since we can override the global `url` setting with a per-entry `url=true`.

This is a change only to the drivers, so I expect it to have little impact on custom styles in general (with the exception that `doi` and `eprint` info might be shown for `@unpublished` where it wasn't shown before). A short (and by no means thorough) search suggests

* @alex-ball's `biblatex-bath` would be affected (specifically https://github.com/alex-ball/bathbib/blob/e2aad742e1d2a67bf201515b883287990161061d/biblatex/biblatex-bath.dtx#L3157-L3161), because it patches the bit of the driver that would be modified

## Add `eid` to more entry types (#847)

The growing importance of electronic publishing may well mean that we are going to see more entries that are identified via a page range (in the `pages` field) any more, but instead have a paper number or non-standard "page numbers". For `@article`s the paper number can be given in the `eid` field and it seems useful to extend the possibility to use the `eid` field to all other entries that allow for `pages` as well.

https://github.com/plk/biblatex/commit/5782552ee8a8a7d2e020db584822e0e39ba978d8 adds a `\printfield{eid}` to the bibmacro `chapter+pages` so that all entry types with `pages` field can make use of `eid`.

Since this is just a change to a single bibmacro, I don't expect many issues with custom styles even more so since custom styles have not yet tried to use `eid` in entry types other than `@article` (as far as I could see).

### `\bibeidpunct`

While we are at it, it seemed appropriate to add a `\bibeidpunct` in analogy to `pages`' `\bibpagespunct`.

### `pagetotal`

In #847 there were discussions about adding `pagetotal` to those types that support `eid` because it is no longer easily possible to calculate the total number of pages via some mental arithmetic on the `pages` field. Unfortunately, I couldn't see a good way to do that without negative side effects for custom styles, so for now I don't plan on doing that. But if anyone has an idea, please leave a comment in #847.

## `issuetitleaddon` and `journaltitleaddon` (#996)

While all other `title` fields (`title`, `maintitle` and `booktitle`) have a `...addon`, `issuetitle` and `journaltitle` don't have one yet.

`issuetitleaddon` and `journaltitleaddon` would be fairly specialised but I can imagine some use cases (`issuetitleaddon` may be useful when the special issue of a journal has a title that doesn't make it clear that it is a special issue and one wants to emphasise that; `jornaltitleaddon` may be useful in case a journal has been renamed or if the journal name isn't unique). Since `issuetitleaddon` and `journaltitleaddon` are analogous to the other `...titleaddon` fields the semantics are clear and it is extremely easy to add them to existing bibmacros (there is a clear place in `issue` and `journal`).

Again, since these are changes to rarely modified bibmacros I don't expect many issues with custom styles.


moewew:
In case anyone is interested in testing these changes. [biblatex-dev-3-15-gh1000.zip](https://github.com/plk/biblatex/files/4493680/biblatex-dev-3-15-gh1000.zip) contains the `tex/latex/biblatex` bit of the TDS tree for `biblatex`. The development version of Biber is not strictly required to test these changes (some other things might explode).

moewew:
There haven't been any complaints in the last four weeks, so I went ahead and merged this. There is still time to fix things if this change broke anything and we still have time to consider additional changes.

The only comment we got was over on #847 on `pagetotal`. I'm still not quite sure what to do there.

moewew:
The changes to the `journal` bibmacro https://github.com/plk/biblatex/pull/1000/commits/c9182dbf52714514c8145a2b936ab134409ec57f#diff-097c35a9a46b863a1840713991dd9dc1R2848-R2850

https://github.com/plk/biblatex/blob/bdcc1233617c07d832dd711e86ca33dcfa67e0eb/tex/latex/biblatex/biblatex.def#L2838-L2850

lead to undesired punctuation in
```latex
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\begin{document}
\cite{sigfridsson}
\printbibliography
\end{document}
```

because now the `\setunit*` in

https://github.com/plk/biblatex/blob/bdcc1233617c07d832dd711e86ca33dcfa67e0eb/tex/latex/biblatex/bbx/standard.bbx#L784-L797

does not get triggered.

moewew:
One way to solve this issue would be to introduce the otherwise unnecessary test `\iffieldundef{journaltitleaddon}`

```latex
\newbibmacro*{journal}{%
  \ifboolexpr{
    test {\iffieldundef{journaltitle}}
    and
    test {\iffieldundef{journalsubtitle}}
  }
    {}
    {\printtext[journaltitle]{%
       \printfield[titlecase]{journaltitle}%
       \setunit{\subtitlepunct}%
       \printfield[titlecase]{journalsubtitle}}%
     \newunit}%
  \iffieldundef{journaltitleaddon}
    {}
    {\printfield{journaltitleaddon}}}
```


```latex
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\renewbibmacro*{journal}{%
  \ifboolexpr{
    test {\iffieldundef{journaltitle}}
    and
    test {\iffieldundef{journalsubtitle}}
  }
    {}
    {\printtext[journaltitle]{%
       \printfield[titlecase]{journaltitle}%
       \setunit{\subtitlepunct}%
       \printfield[titlecase]{journalsubtitle}}%
     \newunit}%
  \iffieldundef{journaltitleaddon}
    {}
    {\printfield{journaltitleaddon}}}

\begin{document}
\cite{sigfridsson}
\printbibliography
\end{document}
```

moewew:
The `\iffielundef` fix is at https://github.com/plk/biblatex/commit/79ef941aa5a5201762584b379f8e0a9b37e11efe.

Issue #1011, opened 2020-06-06
Should bibstring definitions (and other language changes) live in extras<language>?


moewew:
I wondered about this a while ago in https://github.com/plk/biblatex/issues/964#issuecomment-583103675. I'm not saying we need to or should change anything here, but it somehow appears to me that `\captions<language>` would be much more appropriate for our language switcheroo than `\extras<language>`. In particular if we used `\captions<lang>` then there would be a real difference between `autolang=other,` and `autolang=other*,` which might be nice.

I'd love to hear people's thoughts on this.

jspitz:
FWIW I tend to agree.

moewew:
Thanks for the feedback. Maybe I should just try and implement a "captions" version for testing (if things look OK we can make the `\captions...` version the default and implement an option or another simple switch to get back to the `\extras...` behaviour.)

moewew:
Hrmmm ... one thing I forgot (I thought about this before, so I should write it down now) is that we clean up after ourselves with `\noextras...`. I don't think there is a `\nocaptions...`, so that would be tricky.

In theory we could put our `extras` into `\extras...` and our strings into `\captions...` but that would result in really weird side-effects if only language extras are selected and not captions. So this is out.

jspitz:
> In theory we could put our `extras` into `\extras...` and our strings into `\captions...` 

Yes, this would be the way to go.

> but that would result in really weird side-effects if only language extras are selected and not captions. So this is out.

I cannot follow this

moewew:
> > but that would result in really weird side-effects if only language extras are selected and not captions. So this is out.
> 
> I cannot follow this

`biblatex` does not only have language-dependent strings. There are a few other things like the Oxford comma, date formatting etc. One well-known example is that `french.lbx` does (simplified!)
```
\DeclareBibliographyExtras{%
  \savecommand\mkbibnamefamily
  \protected\def\mkbibnamefamily#1{%
    \textsc{\textnohyphenation{#1}}}%
}
```
and then
```
\UndeclareBibliographyExtras{%
  \restorecommand\mkbibnamefamily
}
```
`\DeclareBibliographyExtras` goes into `\extrasfrench` and `\UndeclareBibliographyExtras` is added to `\noextrasfrench` to clean up any side effects afterwards.

Since there is no `\noextras...` equivalent to for `\captions...` these definitions can't go into `\captions...`  because we can't clean up after ourselves.

[...]

PR #1018, opened 2020-06-21
Write bibstrings and extras to \captions... (#1011)

moewew:
Write bibliography strings and extras to `\captions<language>` instead of `\extras<language>`. See #1011.

This can be changed with the `langhook` option.

With `captions`, the new default, `\UndeclareBibliographyExtras` and `\UndefineBibliographyExtras{<language>}` don't do anything any more. This should be no issue for all standard uses of bibliography extras, since everything is local and kept local by grouping. But it could be problematic if people do weird stuff. In case an issue comes up we can advise people to go back to `langhook=extras,` (or change their setup to local changes if possible) - or we can think of something more sophisticated.

### DIFF
--- a/doc/latex/biblatex/CHANGES.md
+++ b/doc/latex/biblatex/CHANGES.md
@@ -108,7 +108,7 @@
   accepts the values `expl3`, `latex2e` and `auto` (which selects
   `expl3` if the `expl3` version not older than 2020-04-06, this
   is the default).
-  
+
   The `expl3` implementation of the case changer is slightly more
   robust than the home-grown `latex2e` code.
 - The option `bibtexcaseprotection` can be used to turn off the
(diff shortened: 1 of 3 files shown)


plk:
This looks nice. It's a shame we can't split this so that `\DeclareBibliographyStrings` goes to `\captions...` and `\DeclareBibliographyExtras` goes to `\extras...` by default but I see in the other thread that this doesn't map cleanly, at least with babel.

moewew:
Technically both `babel` and `polyglossia` would allow us to separate bibstrings and bibextras between `\captions...` and `\extras...`, but as the examples in #1011 show this would lead to counter-intuitive results (at least that's what I think). I thin `\captions...` is the right choice here, just the `\noextras...` cleanup is a bit of an issue.

moewew:
https://github.com/plk/biblatex/pull/1018/commits/db0b57087611155a88eb28442ac24aa7bab84af1 implements a `\noextras...` cleanup by means of a toggle that we set ourselves to show that we used `\captions...`.

Issue #774, opened 2018-07-22
masters thesis translates to Magisterarbeit in German


padalev:
I have noticed while writing my own thesis that when using the "@mastersthesis" entrytype it translates to "Magisterarbeit" in German. But actually a masters thesis in German is just a Masterarbeit. A Magister is a completely different type of Studies which is quite uncommon nowadays.
I'm quite fine with just using the "@thesis" type, but this can really come as a surprise when a master thesis is suddenly described as "Magisterarbeit".
Solution:
As I see it in the file german.lbx the line
`mathesis         = {{Magisterarbeit}{Magisterarb\adddot}},`
should be changed to
`mathesis         = {{Masterarbeit}{Masterarb\adddot}},`

moewew:
Thank you for the report.

Interestingly, PL is German(-speaking?) and `biblatex` was released at a time where the whole Bologna process was already going (early initial release in 2006, v1.0 in 2010), so I guess he would have known about this...

I agree that certainly nowadays *Masterarbeit* is what you would expect. The question is if we can break backwards compatibility here. Usually we allow localisation files to change on the assumption that the new translations are stylistically superior while still maintaining the intended meaning (or at least being a good translation of the concept the string stands for). Here it is slightly different as you point out...

padalev:
Yes. I have thought about that too.
I think you might have to consider what is the bigger damage that can be done.

I you do not change it to a more modern translation you risk that people still use it accidentally in believe that it would translate to "Masterarbeit" without noticing.

If you change it you might have problems with people still believing it to translate to Magisterarbeit. Though there is no other option for a real "Masterarbeit" and also this change would only affect people that are updating there latex distribution while working on some document. Which is a risk, but I believe that future users that download a fresh latex distribution would rather expect the translation to be "Masterarbeit"
Also the change obviously wouldn't affect already finished and/or published documents.

jspitz:
I think it  would introduce a severe compatibility problem. I have many bib databases consisting of `@mathesis` entries, and I certainly don't expect their output to change without my doing. If I cite something I have recorded as a "Magisterarbeit" (and I do cite them also post-Bologna), I expect that the entry will be correct. Masterarbeit would be incorrect.

moewew:
@jspitz Thank you for confirming my fear.

As unfortunate and awkward or even wrong as it is from today's perspective that `mathesis` is "Magisterarbeit", I don't think we can change things now.

moewew:
[...]

PR #1022, opened 2020-07-08
Correct french translations

dpled:
The french.lbx file indicated that translations needed to be corrected. I corrected some typographical and translation mistakes, basing on the website [orthotypographie.fr](http://www.orthotypographie.fr/intros/index.html), and replaced some latin formulas (_et al._, _sqq._, _op. cit._…) by their french equivalent (because most of french-speaking people don’t understand them).


dbitouze:
Hi. I am one of the previous translators. As this PR concerns only French-speaking people, I suggest we continue in French.

Merci pour ces révisions ! J'avais revu assez en détail les traductions pour les rendre conformes aux [règles typographiques en usage à l'Imprimerie nationale](https://fr.wikipedia.org/wiki/Lexique_des_r%C3%A8gles_typographiques_en_usage_%C3%A0_l%27Imprimerie_nationale) qui, certes comme les ouvrages de Lacroux, est une référence en la matière. Je suis d'accord avec certaines de vos modifications (par exemple concernant les sections et paragraphes), moins avec d'autres :

- les locutions latines (_et al._, _sqq._, _op. cit._, etc.) restent répandues en humanités et ça risquerait de perturber certains de les voir repmlacées par leurs équivalents français (même en maths, _et al._ est très courant) ;
- « `coll.` » vient des règles typographiques en usage à l'IN ;
- « `s\'erie` » → « `collection` » : je ne suis pas convaincu ;
- « in » → « dans » : je ne suis pas convaincu. Et, de toute façon, la plupart demandent à le retirer :) ;
- « cf. » → « voir » : je ne suis pas convaincu : c'est quand même très courant, y compris en français ;
- je ne suis pas complètement convaincu par le remplacement des codes de pays. Si toutefois c'est maintenu, les acronymes ne doivent pas contenir de points : par exemple, « `U\adddot E\adddot` » → « `UE` » ;
- je ne suis pas convaincu par l'abréviation « Fr. » pour « France ».

moewew:
I don't speak French, so sorry if that was already mentioned by @dbitouze.

According to the manual the short form of the country strings `countryde`, `countryfr` etc. should be the official ISO 3166 country names, so I don't think they are supposed to be translated. Quoth the manual

> Country names are localised by using the string country plus the ISO-3166 country code as the key. The short version of the translation should be the ISO-3166 country code. 

and then 

> `countryde` The name ‘Germany’, abbreviated as 'DE'.
> `countryfr` The name ‘France’, abbreviated as 'FR'.

etc. etc. So we probably don't want to change the short forms of the country strings. They are only used for patents anyway, so this shouldn't be much of an issue even if ISO 3166 is despised in French typography.

Regarding Latin terms in the French translation we also have https://github.com/plk/biblatex/issues/555, which you may want to have a look at/comment on.


Finally, please keep in mind that changes to translations have the potential alter people's documents. At least in theory the LaTeX world is quite fond of backwards compatibility and reproducibility: The idea being that a document written 20 years ago (for `biblatex` 10 years would be more realistic) compiles to the same result today. `biblatex` has seen a fair bit of changes over the years, but documents not using any advanced features (for some definition of 'advanced') should still compile without issues to pretty much the same output today (minus some bugs). We have sort of resigned ourselves to the fact that localisation/translation overhauls can break backwards compatibility, but I would like to advocate being as conservative as possible/acceptable with the changes. If some of these changes are only of a stylistic nature and there is no overwhelming majority in support of the change it might be wise to stick with the status quo. Of course we don't want to be the last to stick to old conventions just for the sake of it. There is a balance to be struck here. What we should avoid at any cost, however, is changing *meaning*, cf. https://github.com/plk/biblatex/issues/774.

maieul:
i don't think we should replace latin form (including 'in')with French Form. for the reason explained by Denis Bitouzé. The "verbose-trad" styles, which use them, are, as their name explain, "traditionnal", and so traditionnal forms are Latin forms, not French form.  

Also for "nouvelle collection" in place of "nouvelle serie", that is a very misunderstood. In French,  a "collection" can have multiple "série" inside. 

"et collaborateur" for "and more" is also an error. "Et al." is used when many authors wrote the same text, but that does not mean that the last authors are only "collaborateur". They can have the same status that the first mentionned author. 

Concerning the idem/eadem, I am quite sceptical. Latin use gender form when speaking about people. So, why changing all to the masculine form?

So in my opinion, only the `shorthands` change may be kept. However, also "table" versus "liste" is not obvious. A table (like "table des matières" or " table des figures") implies to have, I think, some reference to an external ressource, like the page number in the "table des matières". So "Liste des abbréviations" is, I think, the best.

moewew:
Just a thought: If there are several competing standards in French typography, it would be possibly to write separate `.lbx` files for each standard (using inheritance you don't have to copy the complete `.lbx` file, it would be enough to list the differences). You could then upload that as an independent project to CTAN (`biblatex-lbx-french` or some such), users could then choose different French styles via

```latex
\DeclareLanguageMapping{french}{french-lni}
```
or
```latex
\DeclareLanguageMapping{french}{french-orthotypo}
```
(Of course you could build a higher-level interface around `\DeclareLanguageMapping`.)

dpled:
@moewew It seems to be a good idea, because J.-P. Lacroux recommends to replace latin by french, but not the I.N.… The new package could be an alternative for people who don’t want to have latin forms. I propose to maintain latin forms in the default lbx, and I’ll upload a package on CTAN with Lacroux’s recommendations, so people can choose .

But there are some changes that must be merged in the main code :
− In french, "§" doesn’t mean "section" but "paragraph" ;
− "Mars" shouldn’t be abbreviated, and the short for "juillet" is normally "juill." (because of french abbreviation rules : the break must be between a vowel and a consonant, and − normally − not at the very end of a word).

@dbitouze J’ai remplacé « coll. » par « collab » pour ne pas confondre « collaborateur » et « collection »… Mais si je fais un nouveau paquet avec les recommandations de J.-P. Lacroux, je le changerai moi-même dans ce paquet. J’ai remplacé FR par Fr. car je l’avais déjà vu ailleurs − sur les indications de pays d’origine de films notamment −, et parce que si on mettait « All. » pour l’Allemagne, les majuscules devenaient choquantes à mes yeux… quant à U.E., c’était encore pour suivre les recommandations de Lacroux (sur les points abbréviatifs). Mais puisque les abbréviations doivent être les codes ISO, le problème ne se pose plus… 

@maieul Merci pour vos conseils conçernant les séries et sur l’usage d’« andmore », je les appliquerai dans ma version de la traduction. J’avais mis « Table des abbréviations » car c’est l’intitulé de la rubrique sur le site d’Orthotypo, mais il est vrai que dans les tables (des matières, des illustrations…), on a toujours une page de mentionnée. Et pour idem, c’est parce que je le considère comme passé dans la langue française − car on l’emploie dans le langage courant, comme etcétéra, via et consorts −, mais seulement sous sa forme masculine. Dans mon souci de supprimer les locutions latines, je m’étais dit qu’un francophone non latiniste ne comprendrait pas ce que veut dire « _eadem_ ». Mais je réserve cette modification à la version « Orthotypo » de la traduction, que l’on pourra installer − ou pas… − depuis le CTAN.

moewew:
OK. Let's do it like this: We close this pull request and you can open a new pull request with the 'uncontroversial' changes that are not a matter of style. The more radical changes can then go into your new package to minimise disruption in existing documents.

If there is anything I can help with, don't hesitate to give me a shout.

dpled:
Ok, thanks for your advice. I close the P.R.

PR #1027, opened 2020-07-17
Improve french.lbx

dpled:
In accordance to #1022, I’ve modified the french.lbx file. Changes are :
− "§" have been corrected (in french it means "paragraph", not "section") ;
− "Liste des sigles" has been replaced by "Liste des abréviations" ;
− Abbreviations for "mars" and "juillet" have been corrected ;
− As mentioned in #552 and #555, French texts should always use italic for latin forms (it’s recommended by both Imprimerie nationale and Jean-Pierre Lacroux). As suggested by @moewew in #555, I use _\mkbibemph_ to do so.

Issue #899, opened 2019-06-17
Add an option to make et al. italic


veraverto:
It would be great to add an option to biblatex which writes the _et al._ in italic text. This would make a manual modification of style files unnecessary.

moewew:
Thank you for that suggestion.

I'm not too sure if we really need an *option* here. But something like
```latex
\documentclass[british]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=authoryear, backend=biber]{biblatex}

\newcommand*{\mkbibetal}[1]{#1}

\renewbibmacro*{name:andothers}{%
  \ifboolexpr{
    test {\ifnumequal{\value{listcount}}{\value{liststop}}}
    and
    test \ifmorenames
  }
    {\ifnumgreater{\value{liststop}}{1}
       {\finalandcomma}
       {}%
     \printdelim{andothersdelim}\bibstring[\mkbibetal]{andothers}}
    {}}

\addbibresource{biblatex-examples.bib}

\begin{document}
\cite{sigfridsson,aksin}
\printbibliography
\end{document}
```
might work.

Of course you can always just add the italics to the bibstring.
```latex
\documentclass[british]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=authoryear, backend=biber]{biblatex}

\DefineBibliographyStrings{english}{
  andothers = {\mkbibemph{et\addabbrvspace al\adddot}}
}

\addbibresource{biblatex-examples.bib}

\begin{document}
\cite{sigfridsson,aksin}
\printbibliography
\end{document}
```
That is also not that much work.

Related https://github.com/plk/biblatex/issues/555, discussion in https://github.com/plk/biblatex/pull/552 as well as several TeX.SX questions (a recent one touching upon this subject would be https://tex.stackexchange.com/q/495844/35864, the big one https://tex.stackexchange.com/q/40798/35864, there is also https://tex.stackexchange.com/q/352114/35864, some of these have some more discussions in the comments).

veraverto:
Thank you for your advice! I see that it is possible to change the bibstring quite easily and I am unsure if an option for this is useful in general. I thought a distinct option would make Biblatex more user friendly.

moewew:
[...]

PR #1028, opened 2020-07-18
Implement bibliography string (formatting) classes (#555, #1027)

moewew:
Inspired by #555, #1027, etc. This should make it easier to apply formatting to bibstrings. While it is tempting to just include formatting commands like `\mkbibemph` directly in the strings, it doesn't always yield best results as demonstrated in
```latex
\documentclass[french]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=authortitle-ibid, backend=biber]{biblatex}

\DefineBibliographyStrings{french}{
  andothers = {\mkbibemph{et\addabbrvspace al\adddot}},
  ibidem    = {\mkbibemph{ibid\adddot}},
}

\begin{document}
Lorem \autocite{sigfridsson}
ipsum \autocite{sigfridsson}
dolor \autocite{aksin}
\printbibliography
\end{document}
```
The "ibid." in the second footnotes fails to be capitalised properly
screenshot of the footnotes from the MWE: in particular footnote 2 shows "ibid." in lowercase.
> 2. *ibid.*

---

With this commit we can define a new class for Latin terms and apply additional formatting  (in the right place) only to those strings.
```latex
\documentclass[british,french]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[style=authortitle-ibid, backend=biber]{biblatex}

\DefineBibliographyStrings{french}{
  andothers = {et\addabbrvspace al\adddot},
  ibidem    = {ibid\adddot},
}

\DefineBibliographyExtras{french}{%
  \DeclareBibstringClass{latin}{andothers,ibidem}%
  \DeclareBibstringClassFormat{latin}{\mkbibemph{#1}}}

\UndefineBibliographyExtras{french}{%
  \UndeclareBibstringClass{latin}}

\begin{document}
Lorem \autocite{sigfridsson}
ipsum \autocite{sigfridsson}
dolor \autocite{aksin}
\printbibliography

\selectlanguage{british}
Lorem \autocite{sigfridsson}
ipsum \autocite{sigfridsson}
dolor \autocite{aksin}
\printbibliography
\end{document}
```
screenshot of the footnotes from the MWE: Footnote 2 now is correctly capitalised to "Ibid.". The English footnotes do not show any of the italic formatting for the French text.

---

This also helps with https://github.com/plk/biblatex/issues/899.

---

@plk (and anyone else who wants to have a go) As always I don't feel particularly confident about the name. Suggestions are welcome. Other comments are of course also appreciated.

### DIFF
--- a/doc/latex/biblatex/CHANGES.md
+++ b/doc/latex/biblatex/CHANGES.md
@@ -129,6 +129,10 @@
   of the `\ifnumerals` and `\ifpages` checks.
 - Deprecate the starred `\DeclareDelimAlias*` in favour of
   `\DeclareDelimAlias` with optional arguments.
+- Added `\DeclareBibstringClass`, `\DeclareBibstringClassFormat` etc.
+  to allow injecting additional formatting for a class of bibstrings.
+  Classes can be defined arbitrarily. These commands are primarily
+  intended for use in localisation modules.
 - **INCOMPATIBLE CHANGE**
   `biblatex` no longer falls back to English for unknown languages.
   Warnings will be triggered if undefined language strings or extras
--- a/doc/latex/biblatex/CHANGES.md
+++ b/doc/latex/biblatex/CHANGES.md
@@ -129,9 +129,9 @@
   of the `\ifnumerals` and `\ifpages` checks.
 - Deprecate the starred `\DeclareDelimAlias*` in favour of
   `\DeclareDelimAlias` with optional arguments.
-- Added `\DeclareBibstringClass`, `\DeclareBibstringClassFormat` etc.
-  to allow injecting additional formatting for a class of bibstrings.
-  Classes can be defined arbitrarily. These commands are primarily
+- Added `\DeclareBibstringSet`, `\DeclareBibstringSetFormat` etc.
+  to allow injecting additional formatting for a set of bibstrings.
+  Sets can be defined arbitrarily. These commands are primarily
   intended for use in localisation modules.
 - **INCOMPATIBLE CHANGE**
   `biblatex` no longer falls back to English for unknown languages.
--- a/doc/latex/biblatex/biblatex.tex
+++ b/doc/latex/biblatex/biblatex.tex
@@ -11366,41 +11366,41 @@ It is possible to add bibliography strings to a bibliography string class to app
 
 \begin{ltxsyntax}
 
-\cmditem{DeclareBibstringClass}{class}{key, \dots}
+\cmditem{DeclareBibstringSet}{setname}{key, \dots}
 
-This commands assigns all \prm{key}s to the bibliography string class \prm{class}.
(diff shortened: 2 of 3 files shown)


plk:
I like the idea - if the "class" is only used to determine the format, how about just:

```
  \DeclareBibstringFormat{andothers,ibidem}{\mkbibemph{#1}}}

```

moewew:
Good point. At least at the moment the classes are only for formatting. And I don't see much use apart from formatting.

The problem with
```latex
\DeclareBibstringFormat{andothers,ibidem}{\mkbibemph{#1}}}
```
would be that allowing a comma-separated list in the first argument would is inconsistent with `\DeclareFieldFormat` and friends.

If we have `\DeclareBibstringFormat` I'd also want a way to define format aliases (so that people can say `ibidem` and `andothers` are `latin` and `latin` is defined as ...). Normally the name for that is built by replacing `Format` with `Alias` (cf. `\DeclareNameFormat` - `\DeclareNameAlias`), this doesn't quite work here, I feel, since `\DeclareBibstringAlias` sounds like it does something different.

Plus there is the whole shenanigans about cleaning up after yourself, which needs some thoughts (currently via `\UndeclareBibstringClass{...}`). For that it appears to be easier to have a named object. Otherwise we'd have to list each string separately.

plk:
How about:

```
\DeclareBibstringSet{latin}{...}
\DeclareBibstringFormat{latin}{...}
```

We already have "set" as a general concept in other places and it really is a set ... I don't think we need to say "SetFormat" as formats will only apply to sets and it'll be documented like that.

moewew:
I like 'set'. https://github.com/plk/biblatex/pull/1028/commits/fd333b1ae9a2923d701f2cbdb064e2f8fcf17ba8 hopefully changes all occurrences of 'class' to 'set'.

I chickened out of going for `\DeclareBibstringFormat` and went for `\DeclareBibstringSetFormat` to make it clear to users that we are formatting all bibstrings in a set and that `\DeclareBibstringFormat` can't natively be used to format a single bibstring. If you feel strongly about dropping the `Set` here, I can reconsider that decision.

moewew:
Merged manually in https://github.com/plk/biblatex/commit/e0816f17ae268b8cde40f91ea3a1587434ad891a, https://github.com/plk/biblatex/commit/091ca1fe84408cc8cf6ccd1f5390a217d7bfbc90, https://github.com/plk/biblatex/commit/f9ed28e0888289d8142a47ba8987ec3f107d49ab, https://github.com/plk/biblatex/commit/4579ccaad74024a951b0e4dfabfb28a32d332583, https://github.com/plk/biblatex/commit/4348f30763b5a1da7ff864353668f37b7c3cbd28.

Issue #1023, opened 2020-07-09
Create a \tiifsavedtitleequal togle


bastien-roucaries:
Hi,

Actually I use this kind of contruct in order to replace identical title between reprint by dash, and I will appreciate a toogle to test if title are identical....
```
\iffieldundef{savedtitleaddon}{%
          \iffieldundef{savedsubtitle}{%
            \iffieldundef{savedtitle}{%
              \iffieldundef{savedmaintitle}{%
                % do nothing
              }%
              {% savedmaintitle defined
                \iffieldsequal{savedmaintitle}{maintitle}{%
                  \restorefield{maintitle}{\dashditto}%
                }{%
                  % do nothing
                }% 
              }%   
            }%
            {% savedtitle defined
              \iffieldundef{savedmaintitle}{%
                % savedmaintitle undefined
                \iffieldsequal{savedtitle}{title}{%
                  \restorefield{title}{\dashditto}%
                }{%
                  % do nothing
                }% 
              }%
              {% savedtitle and savedmaintitle defined
                \ifboolexpr{ test {\iffieldsequal{savedtitle}{title}}%
                  and test {\iffieldsequal{savedmaintitle}{maintitle}}%
                }%
                {%
                  \restorefield{maintitle}{\dashditto}%
                  \clearfield{title}%
                }%
                {%
                     % do nothing
                }%
              }%
            }%
          }%
          {% savedsubtitle defined
            \iffieldundef{savedtitle}{%
              % do nothing subtitle without title error likely
            }%
            {% defined savedtitle
              \iffieldundef{savedmaintitle}{%
                % savedmaintitle undefined
                \ifboolexpr{ test {\iffieldsequal{savedtitle}{title}}%
                  and test {\iffieldsequal{savedsubtitle}{subtitle}}}%
                {
                  \restorefield{title}{\dashditto}%
                  \clearfield{subtitle}%
                }{%
                  % do nothing
                }% 
              }%
              {% savedtitle and savedmaintitle defined
                \ifboolexpr{ test {\iffieldsequal{savedtitle}{title}}%
                  and test {\iffieldsequal{savedmaintitle}{maintitle}}%
                  and test {\iffieldsequal{savedsubtitle}{subtitle}}%
                }%
                {%
                  \restorefield{maintitle}{\dashditto}%
                  \clearfield{title}%
                  \clearfield{subtitle}%
                }%
                {%
                     % do nothing
                }%
              }%
            }%
          }%
        }{% titleaddon defined
[...]

PR #1031, opened 2020-07-26
Implement 'or both undefined' equality tests (#1023)

moewew:
See #1023 

Compare
```latex
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage{babel}
\usepackage{csquotes}

\AtEveryBibitem{%
  \iffieldequals{title}{\obviouslyundefmacro}{T}{F}
  \iffieldequalsorbothundef{title}{\obviouslyundefmacro}{T}{F}
%
  \iffieldequals{subtitle}{\obviouslyundefmacro}{T}{F}
  \iffieldequalsorbothundef{subtitle}{\obviouslyundefmacro}{T}{F}
}

\begin{document}
\autocite{sigfridsson}
\printbibliography
\end{document}
```

The names are still up for discussion.

### DIFF
--- a/doc/latex/biblatex/biblatex.tex
+++ b/doc/latex/biblatex/biblatex.tex
@@ -10195,37 +10195,73 @@ Expands to \prm{true} if the \prm{name list} is undefined, and to \prm{false} ot
 
 Expands to \prm{true} if the values of \prm{field 1} and \prm{field 2} are equal, and to \prm{false} otherwise.
 
+\cmditem{iffieldsequalorbothundef}{field 1}{field 2}{true}{false}
+
+Expands to \prm{true} if the values of \prm{field 1} and \prm{field 2} are equal or if both fields are undefined, and to \prm{false} otherwise.
+
 \cmditem{iflistsequal}{literal list 1}{literal list 2}{true}{false}
 
 Expands to \prm{true} if the values of \prm{literal list 1} and \prm{literal list 2} are equal, and to \prm{false} otherwise.
+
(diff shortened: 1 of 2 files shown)


plk:
I wonder if we should just leave defining such things up to users since we have all the building blocks for this available. They can just make their own macro which does some undef checking first? Not least because this means all of these new macros have to be retrofitted to the multiscript branch ...

moewew:
Mhh, I sort of get what you mean. The implementation of such a test would be a bit awkward for users, but certainly doable
```latex
\newcommand*{\iffieldequalsorbothundef}[2]{%
  \iffieldequals{#1}{#2}
    {\@firstoftwo}
    {\iffieldundef{#1}
       {\ifundef{#2}}
       {\@secondoftwo}}}
```
Looking at the implementation of `\iffieldequals` one can see that this implementation is a bit inefficient since it re-does the undef tests. The tests proposed here would be a bit more efficient, but that is probably not really relevant.

While I was implementing the new tests I actually fixed the implementation of some older tests that had a `FIXME` command and weren't working correctly, so I suggest I separate the fixes into a new commit that I push to DEV. We can then close this PR here and can think about adding the new tests after the 3.15 release.

moewew:
https://github.com/plk/biblatex/commit/2fc7d0632e5a0ca7c7bfe8c4a0dde271c39e4aab is the fix I was talking about.

https://github.com/plk/biblatex/compare/dev...moewew:equalorundef would be the changes of this PR rebased on DEV.

plk:
I fixed those "FIXMEs" in the `multiscript` branch - turned out not to be so easy as I recall.

PR #1060, opened 2020-10-11
Create marathi.lbx

NiranjanTambe:
This file needs an additional redefinition by which word ``URL: " can be translated into Marathi. I tried adding the following line into the code -

```
\DeclareFieldFormat{url}{दुवा\addcolon\space\url{#1}}
```

but that doesn't produce the correct results. Try the following MWE -

```
\begin{filecontents}{test.bib}
    @online{चाचणी,
        title={चाचणी लेख},
        author={तांबे, निरंजन},
        date={२०२०},
        url={https://shabdakosh.marathi.gov.in/}
    }
\end{filecontents}

\documentclass{article}
\usepackage[backend=biber]{biblatex}
\addbibresource{test.bib}
\usepackage{marathi}
%\DeclareFieldFormat{url}{दुवा\addcolon\space\url{#1}} % Uncommenting this works!

\begin{document}
    \cite{चाचणी}
    \printbibliography
\end{document}
```

Where shall I add `\DeclareFieldFormat{url}{दुवा\addcolon\space\url{#1}}` to get `दुवा` instead of `URL`?


hvoss49:
With an up-to-date TL2020 I get an error:

```
[...]
Package biblatex Warning: Language 'marathi' not supported.
(biblatex)                Using dummy definitions on input line 16.


! Package etoolbox Error: Toggle 'blx@lang@captions@english' undefined.
```

Using 

```\documentclass[english]{article}```

then there is no error amd the  redefintion of field URL works for me.

```
\begin{filecontents}[force]{test.bib}
    @online{चाचणी,
        title={चाचणी लेख},
        author={तांबे, निरंजन},
        date={२०२०},
        url={https://shabdakosh.marathi.gov.in/}
    }
\end{filecontents}

\documentclass[english]{article}
\usepackage{biblatex}
\addbibresource{test.bib}
\usepackage{marathi}
\DeclareFieldFormat{url}{दुवा\addcolon\space\url{#1}} % Uncommenting this works!

\begin{document}
    \cite{चाचणी}
    \printbibliography
\end{document}
```

Bildschirmfoto 2020-10-11 um 09 10 20

**Comment by NiranjanTambe** [comment]
@hvoss49 You need `marathi.lbx` in `/texmf-dist/tex/latex/biblatex/`. Kindly copy paste [this code](https://github.com/plk/biblatex/pull/1060/files#diff-b4c4f9c2fb6fdf99fdce9ed3338531e8).

The redefinition works for me too, but I want it to be included in the `marathi.lbx`. Right now I need to add it to my `.tex` file which doesn't make much sense. If I add this definition to `marathi.lbx` instead of my `.tex` file, it stops working :(

**Comment by moewew** [comment]
Is the `.lbx` files supposed to be used with `\usepackage{marathi}` or can it also be used with standard `babel`/`polyglossia`?

`\DeclareFieldFormat` calls can be added to `\DeclareBibliographyExtras`, but they need a `\savefieldformat` and `\restorefieldformat` in `\UndeclareBibliographyExtras`. See https://github.com/plk/biblatex/blob/dev/tex/latex/biblatex/lbx/magyar.lbx for quite a few examples. Generally speaking, changing field formats should be avoided as much as possible, but the current localisation system has a few deficiencies, so at some point changing the field formats may be the only way out.

It appears that `marathi.lbx` doesn't have a `\DeclareBibliographyExtras` section. This is quite unusual, as that section sets up additional language-specific stuff. Most importantly the date formats.

There is some date stuff at the end, but the indentation is mangled and the macros appear to use the same name as the (unusual) helper macros for `american.lbx` and the US date format.

I noticed that sometimes you use `.` at the end of a field, and sometimes `\adddot`. E.g. https://github.com/plk/biblatex/pull/1060/files#diff-b4c4f9c2fb6fdf99fdce9ed3338531e8R321 vs https://github.com/plk/biblatex/pull/1060/files#diff-b4c4f9c2fb6fdf99fdce9ed3338531e8R326. Abbreviation dots should be `\adddot`.

The
```latex
\DeclareRedundantLanguages{english,american}{english,american,british,
canadian,australian,newzealand,USenglish,UKenglish}
```
won't be needed, since you are writing definitions for Marathi, not English.

Since `marathi.lbx` needs UTF-8 input, we'll need something like

https://github.com/plk/biblatex/blob/316598708aa0847ba962e92d146529391db062ff/tex/latex/biblatex/lbx/russian.lbx#L6-L17

NiranjanTambe:
> Is the .lbx files supposed to be used with \usepackage{marathi} or can it also be used with standard babel/polyglossia?

Package `marathi` is nothing but a time-saver. It loads the necessary stuff for typesetting in Marathi like `babel`, `fontspec`, a good font with specific options suited for Xe or LuaLaTeX and sets some additional stretch for the document. One can absolutely avoid the package and use babel/polyglossia instead, but the package does nothing different than that :)

Sorry for the `.` vs. `\adddot`, and `\DeclareRedundantLanguages` mess. I hope I have cleaned it up [now](https://github.com/NiranjanTambe/biblatex/commit/c6e8d6c9c36a36a867f63f958cf712b0269bf202). Please let me know if you think something is still wrong.

I don't know anything about `\DeclareBibliographyExtras`. Is there any standard model that can be followed to create one for Marathi language?

> \DeclareFieldFormat calls can be added to \DeclareBibliographyExtras, but they need a \savefieldformat and \restorefieldformat in \UndeclareBibliographyExtras. See https://github.com/plk/biblatex/blob/dev/tex/latex/biblatex/lbx/magyar.lbx for quite a few examples. Generally speaking, changing field formats should be avoided as much as possible, but the current localisation system has a few deficiencies, so at some point changing the field formats may be the only way out.

I will try to build something with the help of `magyar.lbx` and show what I have done here before committing. Thanks

moewew:
Thanks.

Roughly speaking `\DeclareBibliographyExtras` implements the answers to the questions in https://github.com/plk/biblatex/wiki/Checklist-for-submitting-a-new-localisation-file-(.lbx). But it lets you do a lot more if you have to (see `\DeclareFieldFormat`). You can use almost any language file as a basis, but I wouldn't recommend `english.lbx`/`american.lbx` because of its non-standard date setup. Since you know English https://github.com/plk/biblatex/blob/dev/tex/latex/biblatex/lbx/british.lbx might be a good model for the rough ideas, but if you need more stuff https://github.com/plk/biblatex/blob/dev/tex/latex/biblatex/lbx/magyar.lbx and https://github.com/plk/biblatex/blob/dev/tex/latex/biblatex/lbx/latvian.lbx can show you what's possible.

NiranjanTambe:
@moewew I checked this list, I am no expert in this area, in my opinion it would be better to at least add this file which provides Marathi labels for the necessary English ones and at least "support" Marathi with BibLaTeX. The remaining details can be developed gradually. I will mail some of my friends regarding this and collect some information regarding the typographic standards of Marathi, but this will take time to happen, till then is it possible to add the file as it is? Also can you please let me know if [this](https://github.com/NiranjanTambe/biblatex/blob/c6e8d6c9c36a36a867f63f958cf712b0269bf202/tex/latex/biblatex/lbx/marathi.lbx#L337-L487) part of the code is necessary for the localisation text or not? I couldn't understand for what reason it is there.

moewew:
Mhh, I'd *much* rather have a complete `marathi.lbx` than an incomplete file that supports only some of the features. I don't see a need to rush things.

The linked code in https://github.com/NiranjanTambe/biblatex/blob/c6e8d6c9c36a36a867f63f958cf712b0269bf202/tex/latex/biblatex/lbx/marathi.lbx#L337-L487 shouldn't be necessary and can be removed.

NiranjanTambe:
Thank you so much, I will get back to you soon.

PS - [I've removed that code now.](https://github.com/NiranjanTambe/biblatex/commit/b5ec5b332e85367bafb669a96d0bb0a5554e7c07)

moewew:
Thanks.

If you get stuck with the date formats or anything else on the 'extras' list, don't hesitate to ask for assistance. Sometimes things get tricky and we may have to find workarounds to make things look OK.

---

I just noticed that some strings end in `\adddotspace` instead of `\adddot`. Is that intentional? Ideally strings should not end in punctuation other than an abbreviation dot (`\adddot` or `.\isdot` if you like).

moewew:
Oh, another one:

The country abbreviations in

https://github.com/plk/biblatex/blob/b5ec5b332e85367bafb669a96d0bb0a5554e7c07/tex/latex/biblatex/lbx/marathi.lbx#L316-L321

are empty. Ideally the country abbreviations should be exactly as stated in the documentation and https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/03-localization-keys.tex (i.e. ISO-3166). These strings should only be used for patent numbers, so you don't have to worry Latin letters popping up in all kinds of weird places.

Generally speaking, though, defining certain strings to be empty is dangerous.

NiranjanTambe:
> If you get stuck with the date formats or anything else on the 'extras' list, don't hesitate to ask for assistance. Sometimes things get tricky and we may have to find workarounds to make things look OK.

Sure :)

> I just noticed that some strings end in \adddotspace instead of \adddot. Is that intentional? Ideally strings should not end in punctuation other than an abbreviation dot (\adddot or .\isdot if you like).

Sorry, it is [corrected](https://github.com/NiranjanTambe/biblatex/commit/2f5def41f4f81bce47e7c59f469ab34238f82075) now.

Regarding the country-code thing, do you mean [these](https://github.com/NiranjanTambe/biblatex/blob/patch-1/tex/latex/biblatex/lbx/marathi.lbx#L316-L321) macros should be removed? I will not keep them empty as it is dangerous, but what if these macros are completely removed? Will it default to Latin codes then? Or the user will have a choice to write that particular part manually?

moewew:
The country code macros need not necessarily be removed, but we shouldn't have empty groups/translations.

If that is acceptable, I'd make
```latex
    countryde        = {{जर्मनी}{जर्मनी}},
    countryeu        = {{युरोपीय संघ}{}},
    countryep        = {{युरोपीय महासंघ}{}},
    countryfr        = {{फ्रान्स}{}},
    countryuk        = {{युनायटेड किंगडम}{}},
    countryus        = {{अमेरिकेची संयुक्त संस्थाने}{}},
```
read
```latex
    countryde        = {{जर्मनी}{DE}},
    countryeu        = {{युरोपीय संघ}{EU}},
    countryep        = {{युरोपीय महासंघ}{EP}},
    countryfr        = {{फ्रान्स}{FR}},
    countryuk        = {{युनायटेड किंगडम}{GB}},
    countryus        = {{अमेरिकेची संयुक्त संस्थाने}{US}},
```
If Latin characters are unacceptable in Marathi even in the context of ISO country codes in patent numbers, it's probably safest to just delete these strings and let users deal with them if they need them.

---

When I compiled `marathi.lbx` against https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/03-localization-keys.tex as suggested in https://github.com/plk/biblatex/wiki/Checklist-for-submitting-a-new-localisation-file-(.lbx) I noticed that some translations are missing. Notably some combined 'role strings' (e.g. `editorstranaf`), some date related stuff (e.g. `circa`, `commonera`) and many language names (e.g. `langbrazilian`, `langserbian`)

NiranjanTambe:
> If Latin characters are unacceptable in Marathi even in the context of ISO country codes in patent numbers, it's probably safest to just delete these strings and let users deal with them if they need them.

As far as I know, there are hardly any situations where we use ISO country codes in Marathi. Now the problem with "acceptability" is that there are always differences in opinions as we as a community still lack the standards for various things used in the template `.lbx` file provided. As the creator of this `.lbx` file, I have done things that seem correct **to me**, but there is always a possibility of some other user suggesting another way of writing certain entries. This will only happen when users see something that is done already. Also I would like to repeat that there are very few "standard" usages in such situations and the ones that are already established are listed in the proposed file.

> When I compiled marathi.lbx against https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/03-localization-keys.tex as suggested in https://github.com/plk/biblatex/wiki/Checklist-for-submitting-a-new-localisation-file-(.lbx) I noticed that some translations are missing. Notably some combined 'role strings' (e.g. editorstranaf), some date related stuff (e.g. circa, commonera) and many language names (e.g. langbrazilian, langserbian)

I couldn't find a word for `afterword` in Marathi. I have never came across any `afterword` written in a Marathi book. So I had to exclude all the entries which had afterword.

We don't have language names with respect to their country name. Thus I have only included the language names which are used in Marathi, others are almost never used.

Regarding `circa`, `commonera` I am sorry, that is completely my bad. I have [added them](https://github.com/NiranjanTambe/biblatex/commit/944e980d97a640092c9e089162b882036af8b117) now.

moewew:
> As far as I know, there are hardly any situations where we use ISO country codes in Marathi. Now the problem with "acceptability" is that there are always differences in opinions as we as a community still lack the standards for various things used in the template `.lbx` file provided.

There aren't many places where `biblatex` uses these strings, so this is fortunately not a very pressing issue. But I have to insist on avoiding empty translations in this case. We can either go with
```
    countryde        = {{जर्मनी}{DE}},
    countryeu        = {{युरोपीय संघ}{EU}},
    countryep        = {{युरोपीय महासंघ}{EP}},
    countryfr        = {{फ्रान्स}{FR}},
    countryuk        = {{युनायटेड किंगडम}{GB}},
    countryus        = {{अमेरिकेची संयुक्त संस्थाने}{US}},
```
which would be what the manual documents or we simply don't translate these strings at all (which is also perfectly fine, users will get a warning and will have to think about the output they want).

I'll defer to your judgment on which of the two is 'better' in the absence of any known consensus in the Marathi community (my preference would be to keep the translation and use Latin short forms, but I understand that this may look odd or if there are other concerns about mixing Latin and Devanagari script). Again it is important to remember that these strings are only used for patents and patent identifications in the standard styles. Using localised terms here may actually harm findability/correctness of the bibliography entries.


> I couldn't find a word for `afterword` in Marathi. I have never came across any `afterword` written in a Marathi book. So I had to exclude all the entries which had afterword.

That's absolutely fine. I just wanted to make sure you hadn't taken an outdated `.lbx` file as a basis. (And I wanted to make sure you know about https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/03-localization-keys.tex for testing.)

> We don't have language names with respect to their country name. Thus I have only included the language names which are used in Marathi, others are almost never used.

I'm not quite sure I understand what you are saying, sorry, but again it is fine not to translate certain strings if there is non good translation or you are unsure about a good translation.

NiranjanTambe:
> I'll defer to your judgment on which of the two is 'better' in the absence of any known consensus in the Marathi community (my preference would be to keep the translation and use Latin short forms.

I respect your opinion, but I'm afraid we'll have to agree to disagree on this one. It's very common to use Latin characters as "default" for certain standard international practices. It's more common when a concerned community lacks consensus regarding the same. IMO from here itself this is an unjust practice in a multilingual world. So I would suggest, let's remove that country code snippet entirely for now and see how the time flows. For the sake of users' knowledge, we can add a comment-cum-note at the beginning of the `.lbx` itself. e.g.

```
eg. TODO - country codes. Kindly open an issue for further discussions.
```

Does it sound good to you? :smile:

NiranjanTambe:
> That's absolutely fine. I just wanted to make sure you hadn't taken an outdated .lbx file as a basis.

No. I have used the latest one.

> (And I wanted to make sure you know about https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/03-localization-keys.tex for testing.)

Yes I knew about it, but I somehow forgot to test my `.lbx` against it. Sorry for that. OTOH this test file is really an amazing piece of work for the contributors. I really appreciate the efforts by the maintainers!

> I'm not quite sure I understand what you are saying, sorry, but again it is fine not to translate certain strings if there is non good translation or you are unsure about a good translation.

What I meant is, for us Portuguese is the language of Brazil. We don't call it Brazilian. Similarly for American and some other languages listed there. So I have only added the language-names about which I am 100% sure. I would prefer to omit others as I am not sure about them.

moewew:
> So I would suggest, let's remove that country code snippet for now and see how the time flows. For the sake of users' knowledge, we can add a comment-cum-note at the beginning of the `.lbx`. e.g.
> 
> ```
> eg. TODO - country codes. Kindly open an issue for further discussions.
> ```
> 
> Does it sound good to you? 😄

Yes, I believe removing the translations completely is the sensible thing to do if Latin characters are not acceptable/desirable. I wouldn't bother with the note, though. It is *always* implied that people can file an issue if they are missing any translations.

> What I meant is, for us Portuguese is the language of Brazil. We don't call it Brazilian. Similarly for American and some other languages listed there. So I have only added the language-names about which I am 100% sure. I would prefer to omit others as I am not sure about them.

I see, it's similar in German. I'd not say 'Brasilian' (*Brasilianisch*) is the official language of Brazil, I'd probably say it's '(Brasilian) Portuguese' (*[brasilianisches] Portugiesisch*). Interestingly, though, it doesn't sound jarring to say 'from the Brazilian' (*aus dem Brasilianischen*) when we're talking about translations of books. 

Again, not translating things you are unsure about is the best thing to do (much better than just coming up with an unnatural or bad translation for the sake of it: people get warnings about untranslated strings, but not about badly translated ones).

NiranjanTambe:
Great. [Removed now](https://github.com/NiranjanTambe/biblatex/commit/9d6bb063b0351e0d377f047b06a72803753426ee).

> I see, it's similar in German. I'd not say 'Brasilian' (Brasilianisch) is the official language of Brazil, I'd probably say it's '(Brasilian) Portuguese' ([brasilianisches] Portugiesisch). Interestingly, though, it doesn't sound jarring to say 'from the Brazilian' (aus dem Brasilianischen) when we're talking about translations of books.

I am just guessing, but this might be because of the multilingualism in India. Equating country-names with language-names is very unnatural in Indian languages that I so far know. We have a strong dichotomy in country-names and language-names.

moewew:
Thanks. That should be the translations sorted, I think.

Now we 'just' need to tackle the most important bits of the bibliography extras (i.e. the answers to the questions in https://github.com/plk/biblatex/wiki/Checklist-for-submitting-a-new-localisation-file-(.lbx), cf. https://github.com/plk/biblatex/blob/a4abc28fad769876fbd5e047153f0e77049516b7/tex/latex/biblatex/lbx/german.lbx#L6-L93 or https://github.com/plk/biblatex/blob/a4abc28fad769876fbd5e047153f0e77049516b7/tex/latex/biblatex/lbx/british.lbx#L4-L81, the extras in `english.lbx`/`american.lbx` are a bit tricky because they set up the American date format which needs additional helper macros that may not be needed for Marathi).

NiranjanTambe:
1. * We don't distinguish much between `endash`, `emdash` and normal `-`, but as `-` sign is more accessible on the keyboards, I have so far only seen `-` for showing number range in Marathi literature. Also many Unicode Marathi fonts lack `endash` and `emdash, therefore using `-` seems to be the best choice here.
* No spaces

2. Same as above.
3. We go with the UK way. No comma before the connector `आणि`.
4. Like in English we have (a range of) suffixes attached to the number symbols e.g. `1st - १ला`, `2nd - २रा`, `3rd - ३रा`, `4th - ४था`, `5th - ५वा` and वा continued after number 5, but `वा` also is fragmentable as it is gender sensitive. Possible forms are `वा - Masculine`, `वी - Feminine`, `वे - Neuter`.
5. The word for edition is `आवृत्ती` and it is feminine.
6. The word for series is `मालिका` and it is feminine.
7. * Long date format would be something like `२७ ऑक्टोबर, २०२०` i.e. `27 October, 2020`. 
* If day absent - `ऑक्टोबर २०२०` i.e. `October 2020`
* If month absent `२०२०` i.e. `2020`
* Yes. Leading zeros are used.
* Non-breakable thin spaces.
* We used a word instead of `--`. The word is `ते` for showing range. It must be surrounded by normal and breakable spaces.
* Reduced precision is UK style. [Month name] [year].
8. * Short date format would be something like `२७ ऑक्टो, २०२०` i.e. `27 Oct, 2020`. The list of the month name abbreviations is already present in [`babel-mr.ini`](https://github.com/latex3/babel/blob/master/locale/mr/babel-mr.ini#L71-L82). Note that the dots are not necessary while writing dates.
* If day absent - `ऑक्टो २०२०` i.e. `Oct 2020`
* If month absent `२०२०` i.e. `2020`
* Leading zeros are used.
* Non-breakable thin spaces.
* We used a word instead of `--`. The word is `ते` for showing range. It must be surrounded by normal and breakable spaces.
* UK style writing `27/10/2020` (use of leading zeros when necessary.)
9. Capitalization is not relevant for Marathi.
10. * 12 hr format - `म.उ. २.१५` meaning `2.15 PM`
* 24 hr format - `१४.१५` meaning `14.15`
11. `उन्हाळा २०२०` meaning `Summer 2020`.

NiranjanTambe:
@moewew I have listed the answers of the questions asked [here](https://github.com/plk/biblatex/wiki/Checklist-for-submitting-a-new-localisation-file-(.lbx)#bibliography-extras). Let me know which part of this list requires some coding. If that code is already available somewhere, we can use it directly. Like in this case many definitions follow UK conventions. If you think something is completely new and not found in any other `.ldf` file, let me know about it, I'll try to work on it.

moewew:
Thanks for the answers.

Re **2.** When do you use `-` and when do you use `ते` in date ranges?

Re **4.** How exactly do ordinals work? Are only 1, 2, 3 and 4 different and is *everything* from 5 and above the same? Or is there a recurring pattern (as in English with "22nd"). Does gender only play a role for 5 or also for 1--4?

Re **8.** If dots aren't used in date formatting for month abbreviations would be it be OK to just remove the dots from the bibstrings or are there situations where you would need the dots?

Can you explain in more detail how date ranges are handled in Marathi? For example
```latex
\documentclass{article}
\usepackage{marathi}
\usepackage{csquotes}

\usepackage[backend=biber, style=authoryear]{biblatex}

\begin{filecontents}[overwrite]{\jobname.bib}
@book{elk:a,
  author    = {Anne Elk},
  title     = {A Theory on Brontosauruses},
  date      = {1972-10-23/1972-10-25},
  publisher = {Monthy \& Co.},
  location  = {London},
  url       = {https://example.com/~elk/bronto},
  urldate   = {2020-11-09},
}
@book{elk:b,
  author    = {Anne Elk},
  title     = {A Theory on Brontosauruses},
  date      = {1973-10-23/1973-11-07},
  publisher = {Monthy \& Co.},
  location  = {London},
  url       = {https://example.com/~elk/bronto},
  urldate   = {2020-11-09},
}
@book{elk:c,
  author    = {Anne Elk},
  title     = {A Theory on Brontosauruses},
  date      = {1974-10-23/1975-03-04},
  publisher = {Monthy \& Co.},
  location  = {London},
  url       = {https://example.com/~elk/bronto},
  urldate   = {2020-11-09},
}
\end{filecontents}
\addbibresource{\jobname.bib}

\begin{document}
Lorem \autocite{sigfridsson,elk:a,elk:b,elk:c}

\printbibliography
\end{document}
```

NiranjanTambe:
**Re 2.** Sorry for the confusion. `-` seems good in this case. No need of `ते`.

**Re 4.** 
>Are only 1, 2, 3 and 4 different and is everything from 5 and above the same?

Yes.
>Or is there a recurring pattern (as in English with "22nd")

No.
>Does gender only play a role for 5 or also for 1--4?

Gender plays a role for **all** the numbers. So let us segment every morpheme and see how it works.

| | Original number | Ordinal number (segmented) | Number denoter | Gender marker | Resultant word |
| - | ------------- | ------------- | ------------- | ------------- | -------- |
| 1st | एक (one) | पहि (fir~st~) | ल | ी | पहि+ल+ी = पहिली |
| 2nd | दोन (two) | दुस (seco~nd~) | र | ी | दुस+र+ी = दुसरी |
| 3rd | तीन (three) | तिस (thi~rd~) | र | ी | तिस+र+ी = तिसरी |
| 4th | चार (four | चौ (four~th~) | थ | ी | चौ+थ+ी = चौथी |
| 5th (and above) | पाच (five) | पाच (five~th~) | व | ी | पाच+व+ी = पाचवी |

moewew:
**2** Sorry, what do you mean by 'in this case'? Do you mean that for all the date ranges shown in my last MWE a `-` would be OK or do you mean something else?

**4** What about 0?

NiranjanTambe:
**Re 2** Generally the word `ते` is used for date-ranges, but the current output which uses `-` is also seen many a times. It was my bad to not remember it. So instead of `ते` we can keep `-`.

**Re 8** It feels very unnatural to say `0th` in Marathi, but if asked on a gunpoint, it will still be `वा` (the morpheme that is used for and after 5).

NiranjanTambe:
@moewew 

Any updates on this?

moewew:
I created a pull request with some extras against your branch: https://github.com/NiranjanTambe/biblatex/pull/1. If you could have a look and merge it if it works (and ask for changes if something is wrong), we are probably close to being able to merge here.

NiranjanTambe:
@moewew I am extremely sorry, I didn't get any notification for your MR! I just saw it. I'll check it and get back to you. Thank you so much :smile:

NiranjanTambe:
How to check the ordinals? What should be the macro to call them?

moewew:
You can cite a `@book` with an `edition` field to see the relevant ordinals. Other than that you can always try something like `\AtEveryBibitem{\mkbibmascord{1}--\mkbibfemord{2}--\mkbibneutord{3}}` etc.

NiranjanTambe:
@moewew 

Thanks, I tested the following code.

```
\usepackage{marathi}
\usepackage[backend=biber, style=authoryear]{biblatex}
\begin{filecontents}[overwrite]{\jobname.bib}
@article{गौरी,
  author       = {मोघे, गौरी},
  title        = {तन्त्रशास्त्राचा परिचय},
  journaltitle = {आलोक मराठी नियतकालिक},
  date         = {२०२१},
  series       = {3},
  pages        = {१-५},
}
\end{filecontents}
\addbibresource{\jobname.bib}

\begin{document}
\cite{गौरी}

\printbibliography
\end{document}
```

Comments on the output -

1. In the English output `pp.` is written for denoting the page numbers. In Marathi we need `पृ.` there. We can see this entry [here](https://github.com/moewew/biblatex/blob/MarathiExtras/tex/latex/biblatex/lbx/marathi.lbx#L324-L325).
2. Using Devanagari numeral in the series entry doesn't work. (Related to https://github.com/plk/biber/issues/284)
3. `समाविष्ट: ` (Marathi entry for `inseries`) looks very unnatural in Marathi and actually should be dropped. `In` is only relevant for the `@book` entries and not for the `@article` entries. I tried removing `inseries` from the code, but it prints `In: ` in English which looks worse. Now what to do?

NiranjanTambe:
**Re 1.** Using Arabic numerals (1-5) instead of Devanagari ones (१-५) produces `पृ.` for the page numbers, so this seems to be an issue for Biber and not for BibLaTeX.

One more correction suggested by a friend of mine is by default the inline citations (using author-year style) have a comma after the surname of the author. So in the above minimal example the output should produce `मोघे, २०२१` instead of `मोघे २०२१` (English counterpart would be `Reese, 1958; Homer, 2004`). Can this be done in the bibliography extras?

moewew:
**Re 1** This is definitely partly a `biblatex` problem. Currently only ASCII numerals pass the `\ifnum` tests and its variants (here the *input* is important, if the engine makes ASCII numerals look like Devanagari numerals on output). I'm working on extending this to non-ASCII numerals. I think I have something that works for `\ifnum` tests, but it is probably going to be much harder to make the various range compression features work.

**Re 2** I suggest you post a new issue with MWE at the Biber bugtracker. Adding a comment referring to a different comment at the end of a long discussion is usually a good recipe to have something overlooked.

**Re 3** `inseries` is not used by any of the standard styles as far as I am aware. `in` is used for `@article`s before the journal title. The "in:" before the journal is actually fairly unusual in English and German bibliographies as well, but for those languages it is a style decision and not a localisation decision. If it is absolutely unacceptable in Marathi to have the "in:" here instead of just looking unusual or a bit odd, then we can change the bibliography extras to drop the "in:" for `@article` entries (as we have done for languages like Finnish that heavily rely on case and declension rather than prepositions). I'd only like to do this if absolutely necessary for two reasons: (1) Ideally `.lbx` files should translate strings, not manipulate the style output heavily. I understand that localisation is much more than just translation, but the interface we have at the moment is pretty much translation-focussed. (2) Style changes can only be implemented in bibliography extras, which are usually loaded quite late in the document so that they usually overwrite the loaded custom style and user-provided preamble definitions. This can make for a bad user experience (a classic example are the all caps of `french.lbx` for names, a simple `\renewcommand*{\mkbibnamefamily}[1]{\mkbibbold{#1}}` in the preamble is overwritten by that).

The same considerations as for 3 hold for the comma between author and year in citations. In English (and German) this is a style decision: both forms "work" - standard `biblatex` chose one output, but you can easily switch to the other form. Having the `.lbx` file choose a different form from the standard is possible, but makes for an inconsistent and possibly confusing user experience because we'd have to implement this with bibliography extras, which usually override other preamble definitions.

NiranjanTambe:
**Re 1** Is it acceptable to add some definitions in bibliography extras which can easily be removed once the multiscript support is added to `biblatex`? This doesn't manipulate any style and well it kind of fits in the frame of localisation also. No problem if it is not acceptable, I was just trying to confirm if it fits the policy or not.

**Re 2** Sorry for the confusion, I thought `in` and `inseries` are different precisely for this distinction (`@article` and `@book`), but I was wrong. I don't think it is natural in Marathi to write `In: ` for journals. Of course it won't create any "problems" as it is not mis-communicating anything, so if it is weird to have this style change in the bibliography extras, let's skip it. It is definitely not natural though.

**Re 3** For this I would really insist. It looks really unusual in Marathi to not have a comma after the surname. So please consider this as a requirement.

NiranjanTambe:
With great help from [this](https://github.com/plk/biber/issues/356#issuecomment-739879564) comment the following code produces `पृ. ` before the page numbers, but doesn't produce ordinals before series. Is it because the current definition for printing ordinals use Latin numerals? If yes, can the definition also incorporate Devanagari numerals? Also can `\DeclareNumChars{०,१,२,३,४,५,६,७,८,९}` itself be a part of bibliography extras?

```
\documentclass{article}
\usepackage{marathi}
\usepackage[backend=biber, style=authoryear]{biblatex}
\DeclareNumChars{०,१,२,३,४,५,६,७,८,९}
\begin{filecontents}[overwrite]{\jobname.bib}
@article{गौरी,
  author       = {मोघे, गौरी},
  title        = {तन्त्रशास्त्राचा परिचय},
  journaltitle = {आलोक मराठी नियतकालिक},
  date         = {२०२१},
  series       = {३},
  pages        = {१-५},
}
\end{filecontents}
\addbibresource{\jobname.bib}

\begin{document}
\cite{गौरी}

\printbibliography
\end{document}
```

moewew:
**Re 1** Sure, we could add something now that would be obsoleted by a multiscript version. But I don't think the multiscript version will actually help here, so I'd say we think about a good way to implement this now and keep that. In principle the idea I had in mind works just like your `\DeclareNumChars`. I was just planning on adding Devanagari numerals via an internal interface to avoid interference with user-provided `\DeclareNumChars`.

At the moment I'm not sure whether it would make sense to try to include all non-ASCII numerals (of which there are a *lot*) all the time, or whether we should only include the relevant non-ASCII numerals via the `.lbx` file. Both options seem possible, but the 'all numerals all of the time' approach might have a performance impact as well as a maintenance overhead that might make it unattractive. The `.lbx`-based approach where we only declare the numerals relevant for a language looks slightly more attractive in that regard, but might result in odd effects if you load several languages in a document (you can use Devanagari numerals when you cite a Marathi book, but not when you cite an English book if you use `autolang` language switching with aggressive settings ...).

This should solve the "pp." thingy, the `series` thing is a bit more tricky. There `biblatex` uses `\ifinteger`. That test is only true for numerals for which LaTeX can perform arithmetic operations (things like `\the\numexpr 1-5\relax`, which apparently doesn't work for non-ASCII: `\the\numexpr १-५\relax`), which becomes important later on. This is similar to what I alluded to above with "I think I have something that works for `\ifnum` tests, but it is probably going to be much harder to make the various range compression features work." Range compression also needs arithmetic operations and I don't think those work with non-ASCII numerals at the moment. I don't know if there is a clever workaround here.


**Re 3 & 4** As I say, this all can be done, but it comes at the cost of a much worse and possibly confusing user experience for those who want to modify the exact behaviour we are changing here only for Marathi. That's why I push back on purely stylistic changes. But if this a hard requirement in Marathi because the alternative would be wrong or so outrageously unusual that it would almost be unacceptable, we can of course change the output.

NiranjanTambe:
**Re 1** Loading relevant non-ASCII characters from the concerned `.lbx` file sounds good to me. What help do you need from me to implement this in the extras for Marathi?

Without range compression can we only have series ordinals for 1 digit input? (e.g. `३` outputing `३ रा`)

**Re 3 & 4** As I said in the earlier message let's not touch the `In: `, but the comma is almost always used in the citations, so it would be great to incorporate it in the extras.

moewew:
**1** Alright, it is probably more sustainable to have the required numerals in the `.lbx` file. I'll have a think about the interface and implementation here. 

As mentioned above, this solves only half the problem. Some features like the suffix "रा" for ordinals or range compression (for page ranges) rely on LaTeX being able to do arithmetic on the numbers. If LaTeX can't do that, those features don't work at all (so the ordinal suffix will not work at all, no matter if the input is one or several digits long). I have a very rough idea for a workaround, but it would be a long shot and I see ample chance of the idea not working, so I can't promise anything. I also don't expect to be able to work on it before the weekend. (Of course another workaround could be to hard-code the one digit ordinal suffixes and do a string comparison, but that is hardly a nice idea.)

NiranjanTambe:
Okay, do you think it's a good idea to merge the current file first and then add the relevant part for Devanagari digits?

moewew:
I think I'd like to wait until we have sorted everything out properly, unless there is a very pressing need to merge now. We are planning to release a new version of `biblatex` and Biber soon-ish and I'd rather not ship a half-ready `marathi.lbx` whose interface or output might change.

NiranjanTambe:
Okay. Let me know if any other input is needed from my side. Thanks a lot for your contribution to the bibliography extras for Marathi language! :smile:

moewew:
`biblatex` has the option to compress page ranges like in the following example
```latex
\usepackage[T1]{fontenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[backend=biber, style=authoryear]{biblatex}

\DeclareFieldFormat{postnote}{\mkcomprange[{\mkpageprefix[pagination]}]{#1}}

\begin{document}
Lorem \autocite[120-121]{sigfridsson}

Lorem \autocite[120-134]{sigfridsson}

\printbibliography
\end{document}
```
Is that something that would be interesting for Marathi as well (obviously it would be optional) or is this never done?

moewew:
I have some changes ready that would make
```latex
\documentclass{article}
\usepackage{marathi}
\usepackage[backend=biber, style=authoryear, abbreviate=false]{biblatex}

\DeclareFieldFormat{postnote}{\mkcomprange[{\mkpageprefix[pagination]}]{#1}}

\begin{filecontents}[overwrite]{\jobname.bib}
@article{गौरी,
  author       = {मोघे, गौरी},
  title        = {तन्त्रशास्त्राचा परिचय},
  journaltitle = {आलोक मराठी नियतकालिक},
  date         = {२०२१},
  series       = {५३},
  pages        = {१-५},
}
\end{filecontents}
\addbibresource{\jobname.bib}

\begin{document}
a\cite{गौरी}

b\cite[१-५]{गौरी}

c\cite[१]{गौरी}

d\cite[1]{गौरी}

e\cite[1-5]{गौरी}

f\cite[120-122]{गौरी}

\printbibliography
\end{document}
```
produce
grafik
does that look right to you?

I haven't merged the changes and they require some additions to the `.lbx` file, so I'm afraid you can't easily test the changes yet.

NiranjanTambe:
> Is that something that would be interesting for Marathi as well (obviously it would be optional) or is this never done?

Yes! This definitely looks interesting. I myself want this for a project on which I am currently working. It would be great to have this.

>  Does that look right to you?

Yes. This looks exactly as expected. Is it okay to add `\DeclareFieldFormat{postnote}{\mkcomprange[{\mkpageprefix[pagination]}]{#1}}` by default in the extras?

Also a technical doubt. As `series={५३}` is understood correctly as `53rd series` with your new code, may I know what you have exactly done? Earlier in this thread you have said -

> There biblatex uses \ifinteger. That test is only true for numerals for which LaTeX can perform arithmetic operations (things like \the\numexpr 1-5\relax, which apparently doesn't work for non-ASCII.

So have you managed to make `\ifinteger` understand Devanagari numerals? If yes I would love to know more about it and directly include it in the `marathi` package as it can help users to use LaTeX's powers for different purposes. If you think this is not the appropriate place for this discussion, I won't mind posting an SE question.

Thanks a lot !!

moewew:
> Is it okay to add `\DeclareFieldFormat{postnote}{\mkcomprange[{\mkpageprefix[pagination]}]{#1}}` by default in the extras?

It is technically possible, but it's the same story as with the "in:" and comma between name and year: It would be a departure from the standard style decision. So long as this is a style decision and not a language requirement, I'd rather not change it.

> So have you managed to make `\ifinteger` understand Devanagari numerals?

Not really, no. I cheated a lot. Since Devanagari digits can be mapped one-to-one to the Arabic/US-ASCII digits `0`, `1`, ..., `9` we can use LaTeX's `\uppercase` primitive and `\uccode` to map Devangari numerals to their ASCII-equivalents. Once we have these numbers we can use them as usual.

The way I did it, the conversion is *not* transparent to the macros involved, so we need some modifications to internal (bib)macros and field formats to use the converted numbers, but I don't know if there is a more elegant solution.

```latex
\documentclass{article}
\usepackage{marathi}
\usepackage{etoolbox}
\newcommand*{\getASCII}[2]{%
  \begingroup
  \uccode`\०=`\0
  \uccode`\१=`\1
  \uccode`\२=`\2
  \uccode`\३=`\3
  \uccode`\४=`\4
  \uccode`\५=`\5
  \uccode`\६=`\6
  \uccode`\७=`\7
  \uccode`\८=`\8
  \uccode`\९=`\9
  \uppercase{\def\mytempa{#1}}%
  \edef\mytempa{\endgroup
    \def\noexpand#2{\expandonce{\mytempa}}}
  \mytempa
}

\begin{document}
\getASCII{५३}{\foo}

\show\foo
\end{document}
```

This is definitely not enough to make LaTeX generally accept Devangari numerals where it accepts integers for calculation and I wouldn't know how to do that, so by all means ask on TeX.SX.

NiranjanTambe:
Okay. No problem. It seems good now. Do you think it is ready to merge?

moewew:
The `biblatex` sources are currently in a half-frozen state while we wait for all Biber binaries to be built for the 3.16 release. I'd like to wait until after the release, because `marathi.lbx` needs some modification of core `biblatex` macros and that needs some more testing. Once 3.16 is released I'll merge the required infrastructure changes to the `biblatex` core and then we should be able to merge this PR.

NiranjanTambe:
Okay. No problem.
So for successfully getting Marathi translations in a document, users would require `biber v3.16`, but what about `biblatex`? Which version would they require? I would add it as a note in my upcoming work.

I have a request. Is it possible for you to add the relevant `biblatex` changes for the Devanagari numerals as a PR for `biblatex`? So that I can use the code locally for my upcoming release and add a reference to the relevant PRs (`marathi.lbx` and `biblatex.sty` probably) for the users to see the code. This is of course not an urgent requirement. Only consider it if you think it is convenient.

moewew:
My current plan is to merge all the changes once 3.16 is out. Since Devanagari numerals need support in the `biblatex` core, `marathi.lbx` as it will be at that point will require `biblatex` "3.17" (i.e. the version that is going to be released next after 3.16).

That said, it would probably be possible to make that `marathi.lbx` work with older versions of `biblatex` if you are happy to accept that Devanagari numerals don't work as desired. But *I* don't really want to put any effort into making the `marathi.lbx` we merge here semi-compatible with older versions - we have never done that for other languages and it makes the user experience and the code quite messy.

I will either submit a PR or a self-contained commit referencing this issue with the necessary changes to the `biblatex` core, but I'm planning on only doing that after the dust has settled with the 3.16 release (because I want it to be a commit with a clean history and want to avoid having to resolve tons of merge conflicts back and forth). You can then look at the code and use it. If you intend to transplant the code into a third-party add-on/companion package for `biblatex` to make `marathi.lbx` work with `biblatex` 3.16, I'd urge you to make sure that this add-on package does not break down when `biblatex` "3.17" is finally released. (Experience has shown that people often reuse code so that outdated stuff will continue to be used for a very long time. If the add-on breaks with newer versions of `biblatex` that is probably going to be more frustrating for future users than the fact that only 3.17 and not 3.16 supported Marathi.) I'd also find it problematic if the add-on were to distribute a `marathi.lbx` on CTAN, since that would mean that once "3.17" is released we have two possibly diverging `marathi.lbx`s on CTAN.

NiranjanTambe:
> But _I_ don't really want to put any effort into making the `marathi.lbx` we merge here semi-compatible with older versions - we have never done that for other languages and it makes the user experience and the code quite messy.

Yes. I agree and in fact I did not request you to merge the code used for making Devanagari digits in the bib file interpretable, I just wanted the code to be visible to others which I thought can be achieved with a PR (which can be merged at any point of time of your choice), but of course I have no say in the workflow of `biblatex` and hence I won't force you in this regard.

> I will either submit a PR or a self-contained commit referencing this issue with the necessary changes to the biblatex core, but I'm planning on only doing that after the dust has settled with the 3.16 release (because I want it to be a commit with a clean history and want to avoid having to resolve tons of merge conflicts back and forth). You can then look at the code and use it. If you intend to transplant the code into a third-party add-on/companion package for biblatex to make marathi.lbx work with biblatex 3.16, I'd urge you to make sure that this add-on package does not break down when biblatex "3.17" is finally released. (Experience has shown that people often reuse code so that outdated stuff will continue to be used for a very long time. If the add-on breaks with newer versions of biblatex that is probably going to be more frustrating for future users than the fact that only 3.17 and not 3.16 supported Marathi.) I'd also find it problematic if the add-on were to distribute a marathi.lbx on CTAN, since that would mean that once "3.17" is released we have two possibly diverging marathi.lbxs on CTAN.

I have no intention of releasing `marathi.lbx` ***anywhere*** else. The upcoming work about which I was talking is in fact a free and open source scientific journal in Marathi language which is produced using LaTeX. It's source code is planned to be freely accessible with a copyleft license. My only intention while requesting a publicly visible PR was transparency. (Let's say I use `series={३}` in the code of my journal and my output shows `३रा अंक`, then the user will be surprised as they won't see this in the output that they have compiled. So my workaround for this was to mention the two PRs i.e. one mine and one yours for `biblatex` from which users can also extract the exact code and get the correct output.)

This said, you are completely free to not do this. I won't even expect any justification. After all you are the author and you know the best for your package :smile:

PS - Don't worry! This file won't be visible anywhere else than this PR.

moewew:
Since `biblatex` v3.16 is out now, there is https://github.com/plk/biblatex/pull/1087 with the `biblatex` core code for the non-ASCII digit conversion. I'll integrate these new commands in `marathi.lbx` and will manually merge this pull request in a bit.

moewew:
Please have a look at https://github.com/plk/biblatex/pull/1088: That PR is this PR smashed and rebased for a slightly nicer history.

NiranjanTambe:
Thanks for the help! :)
Best wishes for the release

moewew:
Thanks.

I'll be closing this PR now since #1088 is merged. Some testing of #1088 (or the `dev` branch) would be appreciated to make sure that I haven't messed up the rebasing or forgotten anything important.

I'll also still have to add a note about Marathi support to the manual, so please tell me if you want to be credited and how (i.e. which name you want to see).

Please leave any comments (and also the comment about attribution in the manual) under #1088.

NiranjanTambe:
> Thanks.
I'll be closing this PR now since #1088 is merged.

No problem.

> Some testing of #1088 (or the dev branch) would be appreciated to make sure that I haven't messed up the rebasing or forgotten anything important.

I tried testing, but I am sure I have done something wrong, because I am getting errors with the following code.

```
\documentclass{article}
\usepackage{marathi}
\usepackage[backend=biber, style=authoryear, abbreviate=false]{biblatex}

\DeclareFieldFormat{postnote}{\mkcomprange[{\mkpageprefix[pagination]}]{#1}}

\begin{filecontents}[overwrite]{\jobname.bib}
@article{गौरी,
  author       = {मोघे, गौरी},
  title        = {तन्त्रशास्त्राचा परिचय},
  journaltitle = {आलोक मराठी नियतकालिक},
  date         = {२०२१},
  series       = {५३},
  pages        = {१-५},
}
\end{filecontents}
\addbibresource{\jobname.bib}

\begin{document}
a\cite{गौरी}

b\cite[१-५]{गौरी}

c\cite[१]{गौरी}

d\cite[1]{गौरी}

e\cite[1-5]{गौरी}

f\cite[120-122]{गौरी}

\printbibliography
\end{document}
```

I got -

```
ERROR: Missing = inserted for \ifnum.

--- TeX said ---
<to be read again> 
D
l.4 
  
--- HELP ---
From the .log file...

I was expecting to see `<', `=', or `>'. Didn't.
```

For compiling the `.tex` code I cloned your dev branch and pasted the contents of `/biblatex-dev/tex/latex/biblatex/` in one folder. In that folder I kept the tex code given above and compiled it. Got [this](https://github.com/plk/biblatex/files/5800282/test-bib.log) log. I don't know how to make it compilable. Kindly help me to do further testing.

> I'll also still have to add a note about Marathi support to the manual, so please tell me if you want to be credited and how (i.e. which name you want to see).

My full name is Niranjan Vikas Tambe. If it's not a strict convention of the manual to give the full name, you might simply give my first name. I prefer it that way. (Also ***I*** mostly prefer to write my name in Devanagari script. `i.e. निरंजन`. If possible consider this too.)

moewew:
If you go for the copy-everything-into-a-folder approach you need to tweak the

https://github.com/plk/biblatex/blob/4c0a029acf7c226daacf381957c65df1178cca49/tex/latex/biblatex/biblatex.sty#L14-L16

in `biblatex.sty` to something that is valid like

```latex
\def\abx@date{2021/01/12}
\def\abx@version{3.17-dev}
```

you also need to make sure that all files are in the same folders and not within subfolders.

NiranjanTambe:
> If you go for the copy-everything-into-a-folder approach...
Is there any other way?

moewew:
You could also use the install script, see https://github.com/plk/biblatex#installation-from-github

NiranjanTambe:
Thanks. It worked now.

I tried `date={२०२१-०१-२४} (English version - 2021-01-24)` and expected `२४ जाने., २०२१ (English version - 24 Jan., 2021)` in the list of references, but it failed. Do we still need to put Latin numerals in date entry?

moewew:
Yes and no. Biber reads the date fine, but passes on Devanagari numerals, which don't work in the date macros at the moment. I'll see what I can do.

NiranjanTambe:
Okay. By the way my biber version is `2.15`. Is that a problem? Can you reproduce the problem I explained?

NiranjanTambe:
The following code has two citations with the same year. In this case `biblatex` apparently adds `a` and `b`. See the attached screenshot. It should be replaced by `अ आ इ ई उ ऊ ए ऐ ओ औ अं अः क ख ग घ ङ च छ ज झ ञ ट ठ ड ढ ण त थ द ध न प फ ब भ म य र ल व श ष स ह ळ क्ष ज्ञ` these characters. Note that `अं अः क्ष ज्ञ` in the list provided above are actually more than one Unicode characters, but we need to add this list for this purpose in extras.

```
\documentclass{article}
\usepackage{marathi}
\usepackage[backend=biber, style=authoryear, abbreviate=false]{biblatex}

\DeclareFieldFormat{postnote}{\mkcomprange[{\mkpageprefix[pagination]}]{#1}}

\begin{filecontents}[overwrite]{\jobname.bib}
@article{गौरी,
  author       = {मोघे, गौरी},
  title        = {तन्त्रशास्त्राचा परिचय},
  journaltitle = {आलोक मराठी नियतकालिक},
  date         = {2021-01-24},
  series       = {५३},
  pages        = {१-५},
}
@article{गौरी२,
  author       = {मोघे, गौरी},
  title        = {तन्त्रशास्त्राचा परिचय},
  journaltitle = {आलोक मराठी नियतकालिक},
  date         = {2021-02-24},
  series       = {५४},% I increased it by 1
  pages        = {१-५},
}
\end{filecontents}
\addbibresource{\jobname.bib}

\begin{document}
a\cite{गौरी}

b\cite{गौरी२}
\printbibliography
\end{document}
```
Screenshot

moewew:
I'm locking the conversation here, so we can discuss things further over at https://github.com/plk/biblatex/pull/1088 which was merged (so we don't get confused about what code is actually in `dev` at the moment). I'll post something over there in a sec, @NiranjanTambe it would be appreciated if you could have a look.

PR #1088, opened 2021-01-03
Add Marathi localisation

moewew:
See https://github.com/plk/biblatex/pull/1060. This is just that PR smashed and rebased.

### DIFF
--- a/tex/latex/biblatex/biblatex.sty
+++ b/tex/latex/biblatex/biblatex.sty
@@ -3240,6 +3240,7 @@
   \fi
   \endgroup}
 
+% valid numerals set by the lbx file
 \let\lbx@dononasciinumerals\relax
 
 \def\blx@hook@ifnum{%
--- a/tex/latex/biblatex/lbx/marathi.lbx
+++ b/tex/latex/biblatex/lbx/marathi.lbx
@@ -3,7 +3,7 @@
 % व वाङ्मयविद्या परिभाषा कोश तसेच ग्रंथालयशास्त्र परिभाषा कोश
 % ह्या दोन परिभाषा कोशांचा संदर्भ घेतला आहे.
 % ह्यात सुधारणा सुचवण्याकरिता ह्या कोशांचा आधार जरूर
-% घेण्यात यावा. भाषा संचालनालयाच्या कोशांतील शब्दांचे 
+% घेण्यात यावा. भाषा संचालनालयाच्या कोशांतील शब्दांचे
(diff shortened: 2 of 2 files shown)


moewew:
With https://github.com/plk/biblatex/commit/7f59831b81e2b0c80fe15bf77df524732f848681
```latex
\documentclass{article}
\usepackage{marathi}
\usepackage[backend=biber, style=authoryear, abbreviate=false]{biblatex}

\DeclareFieldFormat{postnote}{\mkcomprange[{\mkpageprefix[pagination]}]{#1}}

\begin{filecontents}[overwrite]{\jobname.bib}
@article{गौरी,
  author       = {मोघे, गौरी},
  title        = {तन्त्रशास्त्राचा परिचय},
  journaltitle = {आलोक मराठी नियतकालिक},
  date         = {२०२१-०१-२४},
  series       = {५३},
  pages        = {१-५},
}
@article{गौरी२,
  author       = {मोघे, गौरी},
  title        = {तन्त्रशास्त्राचा परिचय},
  journaltitle = {आलोक मराठी नियतकालिक},
  date         = {२०२१-०१-२४},
  series       = {५४},% I increased it by 1
  pages        = {१-५},
}
\end{filecontents}
\addbibresource{\jobname.bib}

\begin{document}
a\cite{गौरी}

b\cite{गौरी२}
\printbibliography
\end{document}
```
now gives
grafik
does that look OK?

---

It would also be great if you could tell me of the rendering of your name in the manual looks OK

grafik

I know older versions of LuaLaTeX had problems with some non-Latin scripts (not sure about Indic scripts), so I used HarfBuzz rendering, but I want to be sure we don't mess things up.

NiranjanTambe:
> does that look OK?

Yes! It does. Thank you so much! This work seems to be full now. If I use the following command for installing the latest version of package, will these changes be installed immediately on the local system?

```
sudo obuild/build.sh install 3.17-dev /usr/local/texlive/2020/texmf-dist/
```

and yes my name is correctly rendered. LuaLaTeX with HarfBuzz is personally what I prefer the most for Devanagari typesetting. Even my package `marathi` recommends this combination.

moewew:
Very good. Thanks for checking!

> If I use the following command for installing the latest version of package, will these changes be installed immediately on the local system?

If you run this from your git repository after you pulled in the latest `dev` changes, then yes, though I usually do **not** recommend to install these dev version into `texmf-dist/` of your TeX live. It would be safer to use a TEXMFHOME or TEXMFLOCAL: http://www.texfaq.org/FAQ-privinst. Installing into a directory that is controlled by your TeX system can lead to weird side effects.

PR #1096, opened 2021-01-23
update Romanian.lbx

mafsi:
minor correction


tex/latex/biblatex/lbx/romanian.lbx
```diff
@@ -287,7 +287,7 @@
   byorganizer      = {{organizat de}{org\adddotspace de}},
   byauthor         = {{de}{de}},
   byeditor         = {{editat de}{ed\adddotspace de}},
-  bycompiler       = {{compilat de}{comp\adddotspace de}},
+  bycompiler       = {{alcătuit de}{alcăt\adddotspace de}},
```
moewew, line 290: *alcătuit* for `bycompiler` seems to come from a very different root than *redactor* for `compiler`. Is that the natural way to refer to these roles? (In most other languages `compiler` and `bycompiler` use terms that come from the same root.)

mafsi: We don't have `compiler`. I asked and the work of selecting texts and put it together in an anthology is made by a Redactor.
But the process of selecting texts is written in fromtmatter as: Selected text by, or Anthology "X" , alcătuită (put together, made) by

tex/latex/biblatex/lbx/romanian.lbx
```diff
@@ -401,8 +401,8 @@
   pages            = {{paginile}{pp\adddot}},
   column           = {{coloana}{col\adddot}},
   columns          = {{coloanele}{col\adddot}},
-  line             = {{linia}{l\adddot}},
-  lines            = {{liniile}{ll\adddot}},
+  line             = {{linie}{l\adddot}},
+  lines            = {{linii}{ll\adddot}},
```
moewew, line 405: If I understand correctly those two words use a different case(?) than *coloana*, *coloanele*, *versul*, *versuri* etc. now. On first glance that looks inconsistent.

The strings here (`page`/`pages`, `line`/`lines`, ...) are for pinpoint references in the text: "Sigfridsson 1998, `pages` 30-34" and in the bibliography: "In: Journal of Computational Chemistry 19.4 (1998), `lines` 377–395."

The `pagetotal` strings below are used to show the total number of pages in a book.
"Westport, Conn. and London: Greenwood, 2000. 200 `pagestotal`"

mafsi: So it should be: "Sigfridsson 1998, `paginile` 30-34" and in the bibliography: "In: Journal of Computational Chemistry 19.4 (1998), `liniile` 377–395."

mafsi: > The `pagetotal` strings below are used to show the total number of pages in a book.
> "Westport, Conn. and London: Greenwood, 2000. 200 `pagestotal`"

here should be: "Westport, Conn. and London: Greenwood, 2000. 200 `pagini`

moewew: Does that mean that the state of the translations before this change is correct? In that case could you please roll back the changes (and check the other related translations again just to be sure)?

mafsi: > Does that mean that the state of the translations before this change is correct? In that case could you please roll back the changes (and check the other related translations again just to be sure)?

I reverted in my repo. I'll keep some changes but for compiler I can only match with `redactat de`
We use gender sensitive here: if we consider that the bibliography is showing books, then book is feminin (redactat`ă` / alcătuit`ă`)<- feminin article), but if we consider that is a volume we use neutre (_redactat/aclătuit de_)

mafsi:
Reverted & corrected

tex/latex/biblatex/lbx/romanian.lbx
moewew: I see. It's fine to stick with *alcătuit* if that is the more common way to say this. I only picked up on the apparent inconsistency, but I know nothing about the Romanian language, so obviously I am in no position to recommend anything here.

moewew:
Thanks. Squashed and merged.

If you want to go back on the `compiler` thing because *alcătuit*  is more natural than *redactat* feel free to submit a new pull request.

PR #1098, opened 2021-01-24
Catalan lbx update

sebasvila:
Update the catalan localization. New languages and some improvements.


moewew:
Thank you very much. Squashed and merged.

PR #1113, opened 2021-03-19
update romanian.lbx

mafsi:
@moewew please check if is ok. Also let new strings be as it may be useful for other people.


moewew:
`langbasque`, `frombasque`, `langromanian` look good. But I'd rather not include the new strings `byeditoranco`, `byanco`, `revisededition`. If every `.lbx` files comes with its own set of additional strings that is just going to be extremely confusing. We also won't need `\NewBibliographyString{langromanian}` for the release version.

moewew:
Thanks. Squashed and merged.

I removed the superfluous files ` tex/latex/biblatex/lbx/03-localization-keys.pdf` and ` tex/latex/biblatex/lbx/03-localization-keys.tex`.  It would be great if you could avoid checking them in for the next PR.

mafsi:
Thanks. Sorry for that. My bad!

PR #1114, opened 2021-03-22
Added plural for some translated strings

mafsi:
A book doesn't have only a comment, so in Romanian language is better to have : `edited, translated and comments by`


moewew:
Thanks before I merge this, a small point:

The original English string has 'commentary' in singular (~ a collection of comments) and not 'comments' or 'comment'. In German _Kommentar_ can mean commentary or comment; the word is used in the sense of commentary, so the singular also makes sense there.

Is there a similar dichotomy in Romanian? Could we use the singular in the sense of _commentary_ here or do we need _comment**s**_?

mafsi:
In Romanian language we don't have commentary as collection of commentaries. There is either a single comentary made by someone, or that collection of commentaries and we use plural. There is indeed `comentariu` singular form for a book but is usually associated with a kind of introduction or some explanation  to that book.

Also there are more ambiguous form for `note`: there are notes seen as commentaries in critical editions, there is also sigular form `notă` and this is similar with `comentariu` as in : Note to the second edition (in which publishing house usually say that silently corrected some words that are old and not used anymore in modern language.

PR #1126, opened 2021-04-20
Update romanian.lbx

mafsi:
1. match `comentarii` from #1114 for consistency
2. changed `adnotare` to `note` as a product of an annotator, but keep `adnotator` as a person
3. clean some comments


moewew:
Thanks merged.

PR #1147, opened 2021-05-29
Minor fixes for multiple .lbx files

dfussner:
basque, marathi, polish: add explicit space after `\lbx@fromlang`
brazilian, dutch, galician, portuguese: use `\adddot` instead of period "."
before `\lbx@fromlang`

In languages that don't add an explicit `\space` in the definition of `\lbx@fromlang`, space after the macro will disappear in the output, so I've added \ in these cases. In the four other languages, the period after the abbreviation meant that when a language field is present, the bibstring printed by `\lbx@fromlang` will be capitalized, so I replace the . with `\adddotspace`.

(I caught multiple examples of these errors in Chicago's .lbx files, so I thought I'd look at biblatex's files, too.)


moewew:
Thank you very much.

PR #1152, opened 2021-06-11
Update Romanian.lbx

mafsi:
- added partial translation to year division


moewew:
In English and German it would be a good idea to use an unbreakable space `~` before the numbers, i.e. `Trimestrul~1` instead of `Trimestrul 1` and `Trim\adddot~1` instead of `Trim\adddotspace 1`. Do you think that would make sense for Romanian as well?

mafsi:
> In English and German it would be a good idea to use an unbreakable space `~` before the numbers, i.e. `Trimestrul~1` instead of `Trimestrul 1` and `Trim\adddot~1` instead of `Trim\adddotspace 1`. Do you think that would make sense for Romanian as well?

Yes, is a good idea. Thank you!

moewew:
Alright. Merged and ties added (https://github.com/plk/biblatex/commit/e2f1fbc0cb4b2a55f50bb8401d55d69c127cfed1).

Thank you very much.

PR #1157, opened 2021-07-04
Simplify eprint by using a macro

bastien-roucaries:
Less code, less error and easier custumization of eprint (for instance getting the eprint in orange)

@moewew If you have an idea for the name of the bib macro ?

Note that it is backward compatible

### DIFF
--- a/tex/latex/biblatex/biblatex.def
+++ b/tex/latex/biblatex/biblatex.def
@@ -507,52 +507,41 @@
     {\mkbibordedition{#1}~\bibstring{edition}}
     {#1\isdot}}
 \DeclareFieldFormat{eprint}{%
-  \iffieldundef{eprinttype}
-    {eprint}
-    {\thefield{eprinttype}}%
-  \addcolon\space
-  \ifhyperref
-    {\url{#1}}
-    {\nolinkurl{#1}}%
-  \iffieldundef{eprintclass}
-    {}
-    {\addspace\mkbibparens{\thefield{eprintclass}}}}
+  \usebibmacro{eprinttype+eprint+url+eprintextra:*}
+    {%
+      \iffieldundef{eprinttype}
+        {eprint}
+        {\thefield{eprinttype}}%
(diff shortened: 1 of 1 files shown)


moewew:
Hmmm, I'm not convinced we really need this.

Granted, this idea definitely avoids some repetition of code and that is generally seen as desirable (and I broadly agree with that). But at the same time it obfuscates what is going on with a relatively complex four-argument macro.

Basically my thought process is similar to what I said in https://github.com/moewew/biblatex-ext/issues/27. It is quite probably a niche requirement to change the `eprint` field format. If people need to change it, they may not want to be restricted by the argument structure of the helper macro and may have to redefine everything anyway.

In this instance I prefer the relative clarity of the repeated definitions over the brevity we would gain from adding a helper bibmacro.

----


That said, I gave the code a quick spin with
```latex
\usepackage[T1]{fontenc}
\usepackage{babel}
\usepackage{csquotes}

\usepackage[backend=biber, style=authoryear]{biblatex}
\usepackage{hyperref}

\begin{document}
Lorem \autocite{sigfridsson,baez/article,wassenberg,wilde}

\printbibliography
\end{document}
```
and noticed some issues
grafik
when compared to the status quo
grafik

* The round brackets (parentheses) around the arXiv `eprintclass` shouldn't be there (especially if there is no `eprintclass`, because we don't want an empty pair of brackets, but also if there is one).
* For arXiv entries the links should also include the `eprintclass`.


The second issue could be resolved easily by moving the part for the `eprintclass` into the `\href`.
For the first I think you may want to keep in mind that the argument of `\ifblank` is not expanded, so if you pass it something like `\iffieldundef{eprintclass}{}{<something>}` `\ifblank` will always yield "false" even if `eprintclass` is undefined and we would print nothing.

There is also the minor issue of swapped arguments `#2` and `#3` for `eprinttype+eprint+url+eprintextra:*` in `eprint:googlebooks`.

PR #1161, opened 2021-07-29
Fix delim aliasing with contexts ...

moewew:
... remove support for a list of names in `\DeclareDelimFormat`.

The handling of lists in the first mandatory argument of `\DeclareDelimFormat` was broken anyway and a search on CTAN and the rest of the web suggests only `biblatex-archaeology` would be affected.

Plus, other standard commands like `\DeclareFieldFormat` do not support a list as first mandatory argument, either.


moewew:
I notified the `biblatex-archaeology` author https://github.com/CarlOrff/biblatex-archaeology/issues/32.

Issue #1163, opened 2021-08-04
(Hyper)Link the field and entry names in biblatex documentation


Flowdalic:
The main biblatex documentation (biblatex.pdf) would be much better accessible if field and entry names would link to their description.

moewew:
Yeah, the documentation could need some love. For this change I don't quite see the cost-benefit analysis come out on top, so I don't think *I* would want to put too much time into this, but if you are willing to give this a shot, by all means do!

plk:
I'm going to close this as I can't see much benefit in linking fields and entry types as this isn't that useful to know about as you are looking at the docs: such things are style-dependent anyway.

PR #1164, opened 2021-08-04
(Hyper)Link field and entry names to their description in biblatex's documentation

Flowdalic:
I always found that this is missing and that hyperlinks to the field/entry description would make biblatex's documentation far more accessible (see also #1163). The current state of this PR is just a showcase, where only journaltitle is linked. I first like to hear from you like the idea before I put more work into it. And of course I am happy about any feedback on how to improve this. If you think the idea is sensible, which I would hope, then I am happy to link all the things.

Note that the first two commits are, strictly speaking, unrelated to this change. But I needed those to work on the tex file. You may find them sensible and pick them up, or may drop them.

### DIFF
--- a/doc/latex/biblatex/.gitignore
+++ b/doc/latex/biblatex/.gitignore
@@ -1,0 +1,6 @@
+/*.aux
+/*.fdb_latexmk
+/*.fls
+/*.log
+/*.lot
+/*.toc
--- a/doc/latex/biblatex/Makefile
+++ b/doc/latex/biblatex/Makefile
@@ -1,0 +1,4 @@
+all: biblatex.pdf biblatex-quickstart.pdf
+
+%.pdf: %.tex
+	latexmk -lualatex $<
--- a/doc/latex/biblatex/biblatex.tex
+++ b/doc/latex/biblatex/biblatex.tex
@@ -4,7 +4,7 @@
 \usepackage{hyperref}
 \usepackage{zref-xr}
 
-\setmonofont{Courier New}
+\setmonofont{Liberation Mono}
 \setmainfont[Ligatures=TeX]{Linux Libertine O}
 \setsansfont[Ligatures=TeX]{Linux Biolinum O}
 \newfontfamily\devanagarifont{Shobhika}[%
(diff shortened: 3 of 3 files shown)


moewew:
As hinted at in #1163 I like the idea.

But I would like to be able to stick to a syntax like `\bibfield{<field>}` instead of having to define new user-level macros for each field.

You may be interested in my `biblatex-ext` documentation where I do something like this (not for `\bibfield`s, because `biblatex-ext` doesn't define any new ones, but for `\bibmacro`: `\hyperbibmacro`).

Definition

https://github.com/moewew/biblatex-ext/blob/5b3d15a0792c770ceeeb258974b2fba8ba45ddcf/biblatex-ext.tex#L570-L585

and use

https://github.com/moewew/biblatex-ext/blob/5b3d15a0792c770ceeeb258974b2fba8ba45ddcf/biblatex-ext.tex#L2912-L2915

Flowdalic:
> But I would like to be able to stick to a syntax like `\bibfield{<field>}` instead of having to define new user-level macros for each field.

The problem with e.g. `\bibfield{journaltitle}` vs `\journaltitle{}` is that listings of many fields become then unreadable. Right now we have e.g.

```
\optitem{subtitle, titleaddon, language, number, version, note, location, month, isrn, eid, chapter, pages, pagetotal, addendum, pubstate, doi, eprint, eprintclass, eprinttype, url, urldate}
```

I would prefer

```
\optitem{\subtitle{}, \titleaddon{}, \language{}, \number{}, \version{}, \note{}, \location{}, \month{}, \isrn{}, \eid{}, \chapter{}, \pages{}, \pagetotal{}, \addendum{}, \pubstate{}, \doi{}, \eprint{}, \eprintclass{}, \eprinttype{}, \url{}, \urldate{}}
```

and not

```
\optitem{\bibfield{subtitle}, \bibfield{titleaddon}, \bibfield{language}, \bibfield{number}, \bibfield{version}, \bibfield{note}, \bibfield{location}, \bibfield{month}, \bibfield{isrn}, \bibfield{eid}, \bibfield{chapter}, \bibfield{pages}, \bibfield{pagetotal}, \bibfield{addendum}, \bibfield{pubstate}, \bibfield{doi}, \bibfield{eprint}, \bibfield{eprintclass}, \bibfield{eprinttype}, \bibfield{url}, \bibfield{urldate}}
```

but if this is what you absolutely want, then I'll do the latter.

moewew:
True, but if that is a worry, one could look into parsing the argument of `\optitem` as a comma-separated list and applying `\bibfield` to each list item separately.

Flowdalic:
Good point, I'll try to see what I can come up with.

plk:
Looks like we need to wait on this until something more general is suggested.

PR #1165, opened 2021-08-05
\notecite causes duplicate runs of Biber in automated contexts

thebluepotato:
**TL/DR;** The definition of `\notecite` in `biblatex.def` should not contain `\nocite` in the `loopcode` as it causes duplicate runs of Biber with tools such as `latexmk`

I've been pursuing the ever elusive goal of optimizing LaTeX compile times and found an odd bug with `\notecite`.

Here's a MWEB:
```latex
\documentclass{article}
\usepackage{biblatex}

\addbibresource{\jobname.bib}

@book{key,
  author = {Author, A.},
  year = {2001},
  title = {Title},
  publisher = {Publisher}
}
\end{filecontents}

\begin{document}

This is a test of using notecite \notecite[foo][bar]{key}.

\printbibliography

\end{document}
```

If you run `latexmk` on this file, the following commands are run:
```
latex main
biber main
latex main
biber main
latex main
latex main
```
Biber is run twice because the `.bcf` file generated after the second run of `latex` adds a duplicate `bcf:citekey`, causing `latexmk`'s dependency analysis to believe that Biber has to be run again:

```xml
  <!-- CITATION DATA -->
  <!-- SECTION 0 -->
  <bcf:bibdata section="0">
    <bcf:datasource type="file" datatype="bibtex" glob="false">main.bib</bcf:datasource>
  </bcf:bibdata>
  <bcf:section number="0">
    <bcf:citekey order="1">key</bcf:citekey>
    <bcf:citekey order="2">key</bcf:citekey> <!-- This was added after the second run -->
  </bcf:section>
```
After testing and digging around, I found that the following lines are problematic:

https://github.com/plk/biblatex/blob/9163c931822ffef222d8c533610edabea651c921/tex/latex/biblatex/biblatex.def#L2256-L2261

Since the loopcode is not executed when keys are undefined (i.e., on the first run of `latex`), `\nocite` is not "seen" until the second `latex` run, where `biblatex` then adds a duplicate `bcf:citekey`.

Removing the whole `\nocite` command from the macro fixes this issue and does not seem to change the output in the slightest. TBH, I don't know what use `\nocite` has here, since the initialization of `\notecite` itself already adds the entry to the bibliography.

Running Biber twice (and Latex) can be time-consuming and is redundant here. Moreover, most Latex IDEs probably use dependency analysis anyways (Overleaf is based on `latexmk`). Hence, fixing this might shave off a few seconds (or minutes considering the subsequent runs of Latex).

### DIFF
--- a/doc/latex/biblatex/biblatex.tex
+++ b/doc/latex/biblatex/biblatex.tex
@@ -4746,7 +4746,7 @@ The multicite version of \cmd{avolcite} and \cmd{Avolcite}, respectively.
 \cmditem{notecite}[prenote][postnote]{key}
 \cmditem{Notecite}[prenote][postnote]{key}
 
-These commands print the \prm{prenote} and \prm{postnote} arguments but no citation. Instead, a \cmd{nocite} command is issued for every \prm{key}. This may be useful for authors who incorporate implicit citations in their writing, only giving information not mentioned before in the running text, but who still want to take advantage of the automatic \prm{postnote} formatting and the implicit \cmd{nocite} function. This is a generic, style"=independent citation command. Special citation styles may provide smarter facilities for the same purpose. The capitalized version forces capitalization (note that this is only applicable if the note starts with a command which is sensitive to \biblatex's punctuation tracker).
+These commands print the \prm{prenote} and \prm{postnote} arguments but no citation. This may be useful for authors who incorporate implicit citations in their writing, only giving information not mentioned before in the running text, but who still want to take advantage of the automatic \prm{postnote} formatting and citation tracking. This is a generic, style"=independent citation command. Special citation styles may provide smarter facilities for the same purpose. The capitalized version forces capitalization (note that this is only applicable if the note starts with a command which is sensitive to \biblatex's punctuation tracker).
 
 \cmditem{pnotecite}[prenote][postnote]{key}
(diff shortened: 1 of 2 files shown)


moewew:
Thank you very much for reporting (and offering to fix) this!

Yeah, on first glance I see no reason for the `\nocite` either. A `\...cite` command defined with `\DeclareCiteCommand` should always produce a cite request, so the `\nocite` would be pointless.

I'll play around with this a bit more before merging.

If you want to it would be great if you could change the commit message to something slightly more useful in the title (the body is great, but the title is so generic).

moewew:
One thing that probably needs looking at is the documentation. At the moment it explicitly mentions `\nocite` in the description of `\notecite`. This might be important because `\nocite` sets an internal marker for entries that can be used in Biber sourcemaps, though I think the behaviour of the markers that are currently implemented is such that this change brings no practical difference to the status quo.

thebluepotato:
I've removed the mentions of \nocite in the documentation and clarified that the citation is tracked (not sure about the wording). Indeed, my main use of \notecite was to have invisible implicit citations that would be referenced in the backrefs. I've since created my own `\tagcite` citation command that completely ignores the notes but I think the documentation should still reflect that \notecite is a citation command after all.

moewew:
Looks great. Merged. Thank you very much.

Issue #1141, opened 2021-05-15
Italian string for `urlseen` might produce awkward phrasing


ghost:
Environment: TeXLive 2021 with Biblatex 3.16.
The current italian string for `urlseen` is "visitato il", which works well in most cases, but if the day of the month in which the url was visited is the 11th, the result produced by biblatex is the awkward "visitato il 11...", while the correct phrasing would be "visitato l'11..."
A solution might be to change the translation from "visitato il" to "visitato in data".

Here is a mwe
```
\begin{filecontents}{orari.bib}
@online{example,
  url = {http://example.com},
  urldate = {2021-05-11}
}
\end{filecontents}
\documentclass{article}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}
\usepackage[italian]{babel}
\usepackage[backend=biber]{biblatex}
\addbibresource{orari.bib}
\usepackage{newtxtext,newtxmath}
\begin{document}
\cite{example}
\printbibliography
\end{document}
```

PR #1189, opened 2021-10-04
Rephrase "urlseen" text in "italian.lbx" (#1141)

matteocoder:
This change was made to avoid the possibility of producing mistakes such as "visitato il 11" or "visitato il 8".

Edit: to be clearer, if you want to keep the "visitato il" expression, then it should change the article from "il" to " l' " to produce the correct sentence when you have a date whose day contains the number 11 ("undici") or 8 ("otto").
Changing the expression to "visitato il giorno" sidesteps this problem, since it will produce the correct sentence no matter which date is appended to the former ("visitato il giorno 11" and "visitato il giorno 8" are both correct).

### DIFF
--- a/tex/latex/biblatex/lbx/italian.lbx
+++ b/tex/latex/biblatex/lbx/italian.lbx
@@ -338,7 +338,7 @@
   version          = {{versione}{ver\adddot}},
   url              = {{indirizzo}{indirizzo}},
 % urlfrom          = {{}{}},% FIXME: missing
-  urlseen          = {{visitato il}{visitato il}},
+  urlseen          = {{visitato il giorno}{visitato il giorno}},
 % inpreparation    = {{}{}},% FIXME: missing
 % submitted        = {{}{}},% FIXME: missing
 % forthcoming      = {{}{}},% FIXME: missing


matteocoder:
@moewew what are your thoughts on this pull request?

moewew:
Sorry, forgot about this.

Since I know nothing about Italian, I had hoped some of our contributors for `italian.lbx` would share their views. According to the documentation that are @eg9 and another person whose GitHub user name I don't know. @fargiolas also contributed (without a mention in the docs). Thoughts?

---

In theory we could also try to detect the 8 & 11 case and change the output accordingly, though that would probably happen via a custom field format, so would have to be overwritten in a `\DefineBibliographyExtras` (much like we did lots of dark magic for Hungarian), which can confuse people.

matteocoder:
@moewew thanks for your reply. In general, in Italian, if a masculine word begins with a vowel, we should use the definite article "lo". Numbers are all masculine, and 1 (uno), 8 (otto) and 11 (undici) start with a vowel, so we use "lo", but the "o" is dropped to prevent the two adjacent vowels from clashing. Interestingly, in terms of dates, this does not happen for the first day if the month, since people just pronounce it as "primo" (for example "il primo agosto" means "the first of August").
Another workaround would be to force a date format which puts the year as the first element, such as ISO 8601, but that would defer the problem until the year 8000.

plk:
Can we revisit this and perhaps decide on a solution or is this change good enough for now?

matteocoder:
@plk I'm still here. The two people who @moewew contacted for confirmation haven't responded in months, so is their input still necessary to come to a decision?

moewew:
I'd like to be careful with changes like these because they can impact many documents (in a technically non-backwards compatible way), that's why I prefer feedback from other language contributors as well. But I guess if no one raises any doubts about this, we can go ahead with it.

Just to get a feeling of this change: Is "visitato il giorno `<date>`" more awkward or unusual than "visitato il `<date>`"? As I say, we can try to implement something that switches between "il" and "l'", but that might need a bit more work.

matteocoder:
@moewew I understand your point regarding being conservative when it comes to changes like these.

"visitato il giorno" is no more awkward or unusual than "visitato il". Admittedly, the change I propose is just a workaround, whereas the automatic switch mechanism you propose is a more elegant solution.

In the end, I am okay even if this pull request is rejected and the issue is solved in a different manner.

How would the automatic "il"/"l'" switching work under the hood?

moewew:
> How would the automatic "il"/"l'" switching work under the hood?

No idea yet ;-). We'd have to grab the day of the `urldate` and change the string (or use a different string) depending on that.

I probably won't have time to look into this in more detail until the summer (unless it turns out this is very easy).

fargiolas:
sorry for the delay, not sure I have any qualification to say anything about the issue beside being Italian :-)

"visitato il giorno" seems good to me, maybe a bit too verbose but guess that's ok.

Are we sure there aren't any corner cases where this would be wrong? is the date guaranteed to be a dd-mm-yyyy date or can it also be just mm-yyyy or something similar? "visitato in data" may be slightly better if dates without the day field are allowed.

agreed that handling the article correctly as per italian grammar would be the ideal solution

matteocoder:
@fargiolas @moewew "visitato in data" is another good candidate as well.

moewew:
> Are we sure there aren't any corner cases where this would be wrong? is the date guaranteed to be a dd-mm-yyyy date or can it also be just mm-yyyy or something similar? "visitato in data" may be slightly better if dates without the day field are allowed.

In theory `urldate` could contain all kinds of dates allowed in `biblatex` (date ranges, only month-year, ...). In practice I don't think we need to worry about anything other than `yyyy-mm-dd` input. Even the English and German localisations (the "model" localisations) don't really work for anything other than `yyyy-mm-dd`. If you give the URL access date that is anything other than `yyyy-mm-dd` (plus possibly time), you are asking for trouble.

Since I don't speak Italian, I'll have to defer to you on this one, assuming we need not worry about dates other than `yyyy-mm-dd` would "visitato il giorno" or "visitato in data" be more natural?

matteo339:
> Since I don't speak Italian, I'll have to defer to you on this one, assuming we need not worry about dates other than yyyy-mm-dd would "visitato il giorno" or "visitato in data" be more natural?

"visitato in data" or "visitato il giorno" are both valid. It's just a matter of taste.

Matteo

matteocoder:
Sorry to ping again, but are there any updates about this PR?

fargiolas:
Not sure if this was waiting for my take, but I agree that both are fine, please go ahead!

moewew:
Alright. Since people agree that "visitato il giorno" is fine, let's go with that.

Issue #374, opened 2016-03-10
Set up a full set of tests


josephwright:
As discussed in #372, a full set of tests are needed. I'll look at this: first making sure we can run things appropriately, then looking for assistance. If necessary I'll adjust `l3build` to make this possible.

u-fischer:
The newest biblatex broke also the biblatex code of tex4ht (http://puszcza.gnu.org.ua/bugs/?273). Imho if would be good to add also some test code for tex4ht to the test suite. Michal could perhaps help to make it.

plk:
That would be good. I think most biblatex releases break tex4ht.

josephwright:
I'm going to look at the code merge first then sort this: give me a little while (say until Easter)!

plk:
No rush. Likely 3.4 will be out before then and there is one large-ish change to prefixnumber handling to address some user issues and limitations (see CHANGES file which I will make much more prominent in the wake of 3.3 …)

PK

> On 12 Mar 2016, at 10:49 am, Joseph Wright <notifications@github.com> wrote:
> 
> I'm going to look at the code merge first then sort this: give me a little while (say until Easter)!
> 
> —
> Reply to this email directly or view it on GitHub.
> 

--
Dr Philip Kime

u-fischer:
Take your time. I only added the comment because I say the tex4ht message and made the connection and didn't want to forget it.

michal-h21:
Yes, tex4ht support generally breaks with every release, because it redefines lot of internal macros (of `biblatex` and of LaTeX itself). It got little bit messy because we need to keep support also for older biblatex versions. 

One reason why it is hard for us to hack in the current version, is some stuff which is called in `\AtEndPreamble` in `biblatex_.sty`, namely patching of `\MakeUppercase` and `\MakeLowercase` commands, which fails because they are patched by `tex4ht` as well. 

I can hopefully help with testing the `tex4ht` output.

plk:
If we can help at all by putting code into biblatex to ease the tex4ht releases, I'm all for it - tex4ht is an important package for a lot of academics who have to submit in other formats and biblatex is not as journal friendly as bibtex.

plk:
Just so you know @josephwright - there is a meagre start to this already in the `build` subdir of the biblatex dist. I had implemented some tests for the range macros as a start.

LukasCBossert:
Just out of curiosity. Is the test suite for `biblatex` working? Is it accessible or can it be modified to test styles as well?

[...]

PR #1207, opened 2022-02-23
L3build

josephwright:
This PR enables `l3build` for testing, etc. At present, GitHub Actions is set up only for testing, as the documentation typesetting needs a system font.

The current tests are simply the examples turned into `.lvt` files. I would suggest that longer-term new tests could be separated out

- Ones that do one LaTeX run and create the `.bcf`/`.aux`, which can then be verified
- Ones that do one or more LaTeX runs and read the `.bbl` file, which can then be tested using `\showoutput`

An approach avoiding running Biber would be faster and would be purely focussed on the LaTeX code.

Running the CTAN target requires manual setting of the `biblatex` version first:
```
l3build tag <version> && l3build ctan
```
I will look at extending `l3build` to deal with that more cleanly once I know that other areas are usable. In particular, I could do with some input on exactly which files go where and whether any rearrangement of the sources would be permitted to help that.


josephwright:
BTW, this would (finally) close https://github.com/plk/biblatex/issues/374 I think

josephwright:
I've not included the dev Biber aspect: that is workable (just have to install in the right place), but at the moment I get test failures that suggest I've got a mismatched Biber/`biblatex`, so I thought that should be handled separately.

josephwright:
I've got a version sorted that works on the dev branch: would you prefer I rebase to 'offer' that one here?

moewew:
Yeah, it would be great to have this work properly on `dev`.

josephwright:
Infrastructure updated: a few test results change as a result and will follow shortly

josephwright:
All working now: I guess I could add `biber` to the cache, but that seems overkill.

Looking forward to hearing what changes are needed :)

PR #1213, opened 2022-03-09
Update danish.lbx

Hen-Rex:
Fixed some double periods, e.g. adding both a period and a \adddot.
Fixed double spacing due to too many blackslashes, such as \space\ and \adddotspace\\
Fixed incorrect or outmoded translations
Added missing translations marked as "FIXME".

### DIFF
--- a/tex/latex/biblatex/lbx/danish.lbx
+++ b/tex/latex/biblatex/lbx/danish.lbx
@@ -75,45 +75,45 @@
 
 \DeclareBibliographyStrings{%
   bibliography     = {{Bibliografi}{Bibliografi}},
-  references       = {{Litteraturhenvisninger}{Litteratur}},
+  references       = {{Litteratur}{Litteratur}},
   shorthands       = {{Forkortelser}{Forkortelser}},
-  editor           = {{redakt\o r}{red\adddot}},
-  editors          = {{redakt\o rer}{red\adddot}},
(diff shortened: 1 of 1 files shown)


plk:
Merge -  very good, thank you. The situation where a localisation string is in the APA `.lbx` and also in the `biblatex` standard `.lbx` should only occur when the APA string is different from the default `biblatex` string. There are situations like that in english and for other languages, I have to rely on the APA-like practices of other languages that are submitted by contributors familiar with the language.

Hen-Rex (commented):
Changed capitalisation errors in bathesis, mathesis, candthesis and phdthesis.

moewew:
Is there a specific reason why there are now curly braces around `\ae`, `\aa` and `\o`? I expected those not to be necessary. In `.bib` files compiled with BibTeX the braces are needed in order to get "correct" sorting. But in normal text I don't think they are useful. In fact they might stop kerning from working as expected (though I must admit I don't know if `\ae`, `\aa`, `\o` have any kerning info at all, especially when they are macros)

Hen-Rex:
It's just the way we usually do it with the letters æ ø å. And also `redakt{\o}r` is more legible and less prone to spelling errors than `redakt\o er` in my opinion. I tested it and compiling works fine. I did not test if bolded, italics, etc. works as intended though it is pretty standard to format it this way with the special three Danish letters.

PR #1215, opened 2022-03-09
Update danish.lbx - missing squirly bracket causing compile error

Hen-Rex:
- Added a missing squirly bracket that caused a compile error.
- Additionally, it now gives errors for some of the newly added items that were commented out before (Package xkeyval Error: `Q1' undefined in families `blx@lbx'.), etc. 
- I presume that these errors stem from an incompatibility between my version of biblatex v. 3.16 and this dev branch of biblatex-apa and will be fixed once the final version rolls out.

### DIFF
--- a/tex/latex/biblatex/lbx/danish.lbx
+++ b/tex/latex/biblatex/lbx/danish.lbx
@@ -386,7 +386,7 @@
   inseries         = {{i r{\ae}kken}{i rk\adddot}},
   ofseries         = {{fra r{\ae}kken}{fra rk\adddot}},
   number           = {{nummer}{nr\adddot}},
-  numbers          = {{numre}{nr\adddot}},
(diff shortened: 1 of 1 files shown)


moewew:
Yes, the errors you get are because your `biblatex` is older than the file and does not yet know all bibstrings defined there. The current release version is `biblatex` v3.17 (not v3.16), though, and no new strings were added in `dev` since the release. So you should get fewer errors once you update your TeX distribution to a current state.

Note that there is an error due to `numbers`. The string `numbers` does not exists by default. It was commented out before https://github.com/plk/biblatex/pull/1213 (but I don't know why it was there in the first place).

Hen-Rex (commented):
I out-commented the "numbers" variable as this is not needed and wrote a hint to make future reviewers aware.
Contains other uncommitted changes, such as a missing right squirly bracket at "abstract" which causes a compile error.

PR #1257, opened 2022-12-08
Added "eprint:iacreprint" field format for IACR Cryptology ePrint Archive

joachimneu:
Added "eprint:iacreprint" field format for IACR Cryptology ePrint Archive, with recommended citation style according to https://eprint.iacr.org/citation.html


plk:
Looking into this as a specific case of general n2t use.

PR #1260, opened 2022-12-27
Added "eprint:ark"

nkr0:
[ARK](https://en.wikipedia.org/wiki/Archival_Resource_Key) identifiers are used by many organisations. This change adds ARK identifiers in the format, ARK: [53355/cl010066723](https://n2t.net/ark:/53355/cl010066723), and uses [n2t.net](https://n2t.net) for resolving the identifier.


plk:
We're looking into this as a special case of general n2t use.

PR #1293, opened 2023-07-05
Prepare for expl3 case changing change

moewew:
See https://github.com/latex3/latex3/pull/1240

Currently in draft. Will push some doc and polish the comments once details in https://github.com/latex3/latex3/pull/1240 are known.

### DIFF
--- a/tex/latex/biblatex/blx-case-expl3.sty
+++ b/tex/latex/biblatex/blx-case-expl3.sty
@@ -142,10 +142,24 @@
 \cs_set_eq:NN \blx@maketext@uppercase \biblatex_text_uppercase:n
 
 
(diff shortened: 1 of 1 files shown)


josephwright:
Looks reasonable to me. I assume the plan is

- See what exactly is agreed for `expl3`
- Do a `biblatex` (and Biber) release
- Once that is out in TL, I do an `l3kernel` release

so there is basically no point at which a user will be surprised by the `biblatex` behaviour.

moewew:
Exactly that's the idea.

moewew:
This will need to be re-done given that https://github.com/latex3/latex3/pull/1240 is closed and plans have been adjusted.

moewew:
Will open a new PR based on https://github.com/latex3/latex3/pull/1247 when that is approved.

PR #1311, opened 2023-11-05
Typos corrections

antdesros:
English is not my first language, but I'm pretty sure there's something wrong with the word *them* in this sentence. I think it's a typo (the intended word being *then*) but it could be missing words.


plk:
Many thanks for the corrections

PR #1319, opened 2023-11-15
Fix "abbreviation" for refererences

koppor:
In German, the word "Literatur" denotes a **full** bibliography, not the ones being cited (the references). (This was somehow brought up at https://github.com/gi-ev/LNI/issues/94#issuecomment-1537180873).

As far as I understand the intention of the abbreviations, the main use is to shorten the text itself - and not some shorter headings (in general).

Therefore, I propose to use the "right" word "Literaturverzeichnis" also for the abbreviation. (The alternative "Literaturverz." is really, really strange).

For "bibliography", the short form "Literatur" is IMHO OK.

Note that also in Italian, the longer form is also used as "abbreviation". https://github.com/plk/biblatex/blob/a581e08747804b4984fe136fea2ffbd5013754b3/tex/latex/biblatex/lbx/italian.lbx#L91

### DIFF
--- a/tex/latex/biblatex/lbx/german.lbx
+++ b/tex/latex/biblatex/lbx/german.lbx
@@ -95,7 +95,7 @@
 
 \DeclareBibliographyStrings{%
   bibliography     = {{Literaturverzeichnis}{Literatur}},
-  references       = {{Literaturverzeichnis}{Literatur}},
+  references       = {{Literaturverzeichnis}{Literaturverzeichnis}},
   shorthands       = {{Sigelverzeichnis}{Sigel}},
   editor           = {{Herausgeber}{Hrsg\adddot}},
   editors          = {{Herausgeber}{Hrsg\adddot}},


moewew:
Oh, this is a tricky one. I knew that in English some people have strong views on what should be a bibliography and what a "list of references". I can readily believe that something similar carries over to German, where there might be a difference between "Literatur" and "Literaturverzeichnis".

I'm not sure, though, if we can really change this. This is a change to a bibliography string that is extremely widely used and that has huge backwards compatibility implications. (I know many German documents change the bib heading, so this clearly annoys a sizeable number of people. But still...)

As always when German localisations are involved, I'd love to hear @jspitz's thoughts.

jspitz:
The use of _Literatur_ vs. _Literaturverzeichnis_ vs. _Bibliographie_ in German is somewhat arbitrary. I quickly re-checked some works from my own field (Germanic Linguistics), and it seems _Literatur_ is mostly preferred for chapters and papers whereas _Literaturverzeichnis_ is preferred for books (similar to _references_ and _bibliography_ in English), all for the list of _cited_ works (for non-cited works I would probably go for _Auswahlbibliographie_ or _Bibliographie_).

But I also found uses of _Literatur_ in (well-established) books. I think I never saw _Literaturverzeichnis_ in a paper.

Generally I think the use might depend on a field's convention, maybe also on a publisher's preferences. I strongly agree with @moewew that we should not change this also for backwards compatibility reasons.

koppor:
@jspitz Thank you for the input. In computer science, it seems to be the other way round. Thus, we will keep customizing the bibliographic strings for that discipline.

moewew:
@jspitz Thank you very much for the research. So it seems that whichever way we go, some people might be unhappy with - or at least surprised - by the result. So we had better stick with the status quo to avoid a backwards compatibility nightmare.

Issue #1316, opened 2023-11-10
Italian strings to be added


matteofg:
Hi,
the localization file `italian.lbx` is still missing many translations. The following is a possible contribution, loosely based on the translations in the `biblatex-philosophy` package:
```
reviser            = {{revisore}{rev\adddot}},
revisers           = {{revisori}{rev\adddot}},
founder            = {{fondatore}{fond\adddot}},
founders           = {{fondatori}{fond\adddot}},
continuator        = {{continuatore}{cont\adddot}},
continuators       = {{continuatori}{cont\adddot}},
collaborator     = {{collaboratore}{collab\adddot}},
collaborators    = {{collaboratori}{collab\adddot}},
annotators       = {{annotatori}{annot\adddot}},
editortr         = {{curatore e traduttore}{cur\adddotspace e trad\adddot}},
editorstr        = {{curatori e traduttori}{cur\adddotspace e trad\adddot}},
editorco         = {{curatore e commentatore}{cur\adddotspace e comm\adddot}},
editorsco        = {{curatori e commentatori}{cur\adddotspace e comm\adddot}},
editoran         = {{curatore e annotatore}{cur\adddotspace e annot\adddot}},
editorsan        = {{curatori e annotatori}{cur\adddotspace e annot\adddot}},
editorin         = {{curatore e introduzione}{cur\adddotspace e introd\adddot}},
editorsin        = {{curatori e introduzione}{cur\adddotspace e introd\adddot}},
editorfo         = {{curatore e prefazione}{cur\adddotspace e pref\adddot}},
editorsfo        = {{curatori e prefazione}{cur\adddotspace e pref\adddot}},
editoraf         = {{curatore e postfazione}{cur\adddotspace e postf\adddot}},
editorsaf        = {{curatori e postfazione} {cur\adddotspace e postf\adddot}},
editortrco       = {{curatore, traduttore\finalandcomma\ e commentatore}{ed.,\addabbrvspace trad\adddot\finalandcomma\ e comm\adddot}},
editorstrco      = {{curatori, traduttori\finalandcomma\ e commentatori}{eds.,\addabbrvspace trad\adddot\finalandcomma\ e comm\adddot}},
editortran       = {{curatore, traduttore\finalandcomma\ e annotatore}{ed.,\addabbrvspace trad\adddot\finalandcomma\ e annot\adddot}},
[...]

PR #1324, opened 2023-11-26
Add more Italian strings (#1316)

moewew:
Suggestions by @matteo339 in #1316.


moewew:
Will merge as soon as we get the OK by @matteo339 that I didn't mess up any of the suggestions.

PR #1355, opened 2024-04-27
Add \Fvolcite, \Ftvolcite and \ftvolcites to the documentation

PHPirates:
@frankbits noticed that `\Fvolcite`, `\Ftvolcite` and `\ftvolcites` are missing from the documentation. I checked and they appear to be working fine, so I added them.
The exception is `\Ftvolcites` which doesn't work for me, but I can't think of any situation where it would be applicable so not sure if this is an issue.


moewew:
Thank you very much for catching that. Merged.

`\Ftvolcites` didn't work as intended due to a typo. That should be fixed in https://github.com/plk/biblatex/commit/237c4d415d1254e314037688a33531535d1cc34d.

PR #1361, opened 2024-06-08
fix small typos

mbertucci47:
Found with [codespell](https://pypi.org/project/codespell/)


plk:
Thanks for this fix - these typos are indeed an irritation.

PR #1369, opened 2024-07-24
Update Swedish

kess:
Mostly some missing translations but also a couple corrections.


moewew:
Thank you very much. Merged.

PR #1370, opened 2024-07-24
Langs

kess:
I just noticed that the strings in the translation files are not up to date, I wrote a quick script to find the missing entries and added them manually for Swedish.

Are there any tools to (semi)automatically update the strings? If not, I may be able to find some time to do the same (without translations) for the other languages, if it’s wanted that is.


moewew:
Thanks. Merged.

We usually recommend that language contributors test their `.lbx` against https://github.com/plk/biblatex/blob/dev/doc/latex/biblatex/examples/03-localization-keys.tex with the appropriate language setting (see https://github.com/plk/biblatex/wiki/Checklist-for-submitting-a-new-localisation-file-(.lbx) and https://github.com/plk/biblatex?tab=readme-ov-file#translation). That files uses all defined strings and generates warnings for each one that's missing. The missing translations can then be added manually.

PR #1384, opened 2024-09-18
Improving polish.lbx

swasjt:
A new, thoroughly revised, corrected and improved version of the localization file for the Polish language, which aim is to make it consistent with Polish publishing/editorial practice.


moewew:
I've squashed the commits and submitted the PR against `dev` at https://github.com/plk/biblatex/pull/1387. @swasjt and @acuiko does that look right?

swasjt:
Yes, definitely.

PR #1388, opened 2024-09-29
Rewrote estonian.lbx

artjom3729:


moewew:
Thank you very much for getting in touch and contributing localisations to `biblatex`!

@kristiank You previously contributed the Estonian translations. Care to have a look at this and share your thoughts?

Regardless of @kristiank's input, here are a few things I noticed and that I would like to ask about.

* Why exactly does the PR force some strings to be always lowercase with `{}`? Is it wrong to capitalise these words even after a full stop/period? So far previous Estonian language contributors have not mentioned this or complained.
* I notice that the translations make more "aggressive" use of the separation between short and long string than previously. Is that really desirable? I.e. are both the long and short forms equally widely used? I do realise it is a bit weird to ask this since we clearly have the option to switch between long and short forms, but if previous localisation versions stuck with one or the other, maybe there was a reason for it? I'm in particular interested in the short form of the languages. Are those really used and widely recognised?
* Is it widely accepted that abbreviations can be written with or without a dot and is it also accepted that the same document would mix both styles? (Which would be the case here if I understand correctly.)

artjom3729:
Hello there!
1. I didn't know {} would prevent the string from being capitalized even after a full period. According to the rules, these strings should not be capitalized if used within sentences or if they are preceded by a full stop when using ordinal numbers (like "3rd part" - "3. osa").
2. The short forms are not always widely used, however I added them, since they exist and can be used in certain situations. Of course, if the other contributor doesn't agree with the use of these short forms, they can be freely reverted.
3. There are some words that do not use dots (like "ehk" - "e", "inglise keel" - "ingl") and some words that do use a dot (like "toimetaja" - "toim.") because otherwise they would mean a different thing (like "toim" means "a property or appearance of the fabric resulting from the thickness and density of the weft yarns of the fabric"). Some words are okay to be written with or without a dot.
Kind regards

artjom3729:
@moewew Hello there!
It's been a week already. I hope you haven't forgotten.

kristiank:
Hello, and thank you for inviting me to the discussion. Some remarks from me:
1. I'm unable to assess the contextual constraints concerning the lowercasing after ordinal numbers. I leave it up to Artjom to decide whether it is or is not so that the concerned strings will always be preceded by an ordinal number.
2. I agree with Artjom that the short forms exist and should thus be available as an option. The use of short forms for languages is used, and I believe ambiguity is not a problem for readers, as most probably an appropriate context is available to the reader (an example is _ld_, which can be both Lithuanian and Latin but in the context of, e.g., Baltic studies, it is still understandable)
3. Moewew is right in being concerned about mixing styles in the same document. Still, I agree with Artjom's changes, as the use of dots, where they usually are not used, has a connotation to old orthography rules.

I agree with the pull request, but it would be good if Artjom could re-assess point 1 above.

artjom3729:
Hello there!

1. Please refer to this website, which is created by the Estonian Language Institute https://sonaveeb.ee/learn?uilang=et. 
Järgarvude kasutamine
Kui kirjutame järgarvu araabia numbriga, on selle järel punkt.
25. septembril (= kahekümne viiendal septembril)
14. korrusele (= neljateistkümnendale korrusele)
10. reas (= kümnendas reas)

The word begins with a lowercase letter after an ordinal number in all cases.

2. By the way, I made a clear distinction between Latin and Lithuanian: Ladin is _lad_ and Lithuanian is _ld_, as it is clearly permitted https://sonaveeb.ee/search/unif/dlall/dsall/lad/1/est

moewew:
@artjom3729 I wanted to wait for @kristiank's input before moving this further, since I cannot assess the impact of these changes.

@kristiank thank you very much for having a look at this.

* If everything is marked up correctly, `biblatex` should not capitalise strings after a period as an ordinal marker. So if that's the only reason for the `{}` we should get rid of them. If there is a reproducible example where `biblatex` wrongly capitalises such a string, it would be great if you could show it and we can see where the bug hides.
* As for the long/short forms, I'm mainly asking because the default is to show the short form. So changing the translation of the short form from something that was essentially a long form to an aggressively shortened form will have backwards incompatible impact on existing document. Of course this is true for all localisation changes and is expected and in many cases even desired, but in this case long vs short can be thought of as a style decision (where there is no obvious objective right-wrong-distinction). If both of you agree that the abbreviations are used in practice, understandable to most people and desirable, then we have a good reason to change this and accept this part of the PR as is.

artjom3729:
Hello once again!

I decided to remove {} because there's really not so much need in it. As for short forms. I only edited a few of them that are quite too rare.

moewew:
Thank you very much. Merged. If you want to be listed as a language contributor in the documentation, please let me know how you would like to be credited.

artjom3729:
Hello! May you please add me to the BibLaTeX documentation beside other Estonian translators? Thanks.
image

moewew:
Do you want to be credited as Artjom Jemeljanov? (I'm asking specifically, because I can never be sure that the name I get from GitHub has non-ASCII chars right and that this is the name you want to see in the docs.)

artjom3729:
Yes, that's correct.

moewew:
Done in https://github.com/plk/biblatex/commit/e4fd759c6c4b48c69c5ee1c4d667f73cb4178477. Thank you very much again.

Issue #217, opened 2014-03-10
Option to automatically use sequens, sequentis in postnote


ghost:
Biblatex currently has an option to compress ranges of pages in backrefs to use the sequens and sequentis strings where appropriate.
An option to do the same for postnotes in citations seems to make sense.

E.g. 
\autocite[34, 35]{Chomsky1957} would produce Chomsky (1957, pp. 34 sq.)

jspitz:
Here is something I hacked together since a publisher required the use of seq and seqq. This is currently without the starred counterpart and probably can be done much more elegantly (as you can probably see, I just modified the comprange macros until I got what I needed.

Feel free to take and improve as you see fit:
```tex
\newrobustcmd*{\mkseqrange}{%
  \begingroup
    \blx@range@aux\blx@seqrange@i}

\def\blx@seqrange@i[#1][#2]#3{%
  \let\blx@tempa\@empty
  \protected\def\blx@range@out@value{\appto\blx@tempa}%
  \def\blx@range@out@item@process{#2}%
  \let\blx@range@out@delim\blx@range@out@value
  \let\blx@range@split\blx@genrange@split
  \let\blx@range@process\blx@seqrange@check
  \blx@range@chunk{#3}%
  \begingroup
  \edef\blx@tempa{\endgroup
    \unexpanded{#1}{\expandonce\blx@tempa}}%
  \blx@tempa
  \endgroup}

\def\blx@seqrange@check#1#2{%
  \blx@imc@ifinteger{#1}
    {\blx@imc@ifinteger{#2}
       {\@firstoftwo}
       {\@secondoftwo}}
    {\@secondoftwo}
    {\blx@seqrange@comp{#1}{#2}}
    {\begingroup
     \protected@edef\blx@tempc{\endgroup
       \blx@range@out@value{%
         \blx@range@out@item@process{\unexpanded{#1}}%
         \noexpand\bibrangedash
         \blx@range@out@item@process{\unexpanded{#2}}}}%
     \blx@tempc}}

\def\blx@seqrange@comp#1#2{%
  \def\blx@tempb{#1}%
  \def\blx@tempc{#2}%
  \let\blx@tempd\blx@tempc
  \ifnum\c@maxcomprange<\abx@rangeproclimit\relax
    \numdef\blx@tempe\abx@rangeproclimit
  \else
    \numdef\blx@tempe\c@maxcomprange
  \fi
  \blx@tempcntc=\blx@tempe\relax
  \ifnum
[...]

PR #1392, opened 2024-11-01
Introduce option to automatically use sequens, sequentes in postnote (#217)

jspitz:
This entails new formatting commands `\mkseqrange[*]` and `\mkseqornormrange[*]` as well as a new option
`pageseq=none|two+|three+|all+`

The lexic of the latter is adapted from `backrefstyle`.


jspitz:
An afterthought: We could also generalize the option a bit more to provide even more convenience:
* rename proposed macro `\mkseqornormrange` => `\mkautorange`
* make it also consider `\mkcomprange`
* rename proposed option `pageseq` => `citepagerange` with values `normalized` (default), `compressed`, `2sq`, `3sqq`, `allsqq`

I think this would be a useful interface. What do you think?

jspitz:
I have updated my pull request to include the extensions proposed in https://github.com/plk/biblatex/pull/1392#issuecomment-2453305335

jspitz:
BTW as far as I can see the reason for the checks failing is a biblatex/biber version mismatch in `dev` branch.

moewew:
Looks nice. Thank you *very* much. I'd like to squash and merge to keep the history a bit nicer. Would that mess things up hugely on your end?

jspitz:
I think I can deal with it. It's better for me to do a completely new fork after this has been merged anyway in order to start from a clean history.

moewew:
Nice. Merged.

(I guess it is probably easier to work on feature branches in future, then you don't need to worry too much about what's going on on `dev`.)

Issue #1381, opened 2024-09-10
hyphenation warning with language variants and polyglossia


jspitz:
From https://github.com/reutenauer/polyglossia/issues/663:

I have a issue with a document, using the austrian language.
Here is a minimal example file:
```
\documentclass{scrreprt}

\usepackage{polyglossia}
\setdefaultlanguage[variant=austrian]{german}
\usepackage[backend=biber]{biblatex}

\begin{document}
Hier steht Inhalt. Welcher jährlich mehr wird.
\end{document}
```
The run of lualatex gives the following warnings:
```
(/usr/share/texmf-dist/tex/latex/biblatex/lbx/german.lbx
Module polyglossia Warning: Language naustrian not found in language.dat.lua on
 input line 580

Package biblatex Warning: No hyphenation patterns for 'naustrian' on input line
 580.

) (/usr/share/texmf-dist/tex/latex/biblatex/lbx/german.lbx
Module polyglossia Warning: Language naustrian not found in language.dat.lua on
 input line 580

Package biblatex Warning: No hyphenation patterns for 'naustrian' on input line
 580.

)
Module polyglossia Warning: Language naustrian not found in language.dat.lua on
 input line 20

Package biblatex Warning: No hyphenation patterns for 'naustrian' on input line
 20.

)
```
The same file without austrian variant `\setdefaultlanguage{german}` gives no warnings.
Versions are LuaHBTeX 1.18.0, polyglossia 2024/03/07 v2.1, biblatex 2023/03/05 v 3.19

---

AFAICS this is due `\DeclareHyphenationExceptions`. Interestingly, replacing them with `\DefineHyphenationExceptions{ngerman}` does not seem to fix it.

(These exceptions themselves are disputable in the lbx files; but I think these go back to plehmann)

jspitz:
A fix for this would be to use polyglossia's own hyphenation exception macro which takes care for aliases:
```diff
diff --git a/tex/latex/biblatex/biblatex.sty b/tex/latex/biblatex/biblatex.sty
index 6495db6b..1e37447d 100644
--- a/tex/latex/biblatex/biblatex.sty
+++ b/tex/latex/biblatex/biblatex.sty
@@ -562,6 +562,9 @@
     {\ifundef\xpg@ifdefined
        {}
        {\def\blx@ifhyphenationundef#1#2#3{\xpg@ifdefined{#1}{#3}{#2}}}%
+     \ifundef\pghyphenation
+       {}
[...]

PR #1396, opened 2024-11-14
Fix `\DeclareHyphenationExceptions` with polyglossia and language variants (#1381)

jspitz:
As elaborated in #1381 this used `polyglossia`'s own `\pghyphenation` which is aware of babel name aliases.

The PR also removes redundant uses of the same `\DeclareHyphenationExceptions` in the different varieties of German. All of them inherit the exceptions as defined in `german.ldf` and apply them to the right pattern.


moewew:
Thank you very much. Merged.

Issue #1390, opened 2024-10-28
Localizable location?


jspitz:
Many of the (larger) cities where publishers locate have different or differently spelled names in different languages (e.g., Wien -- Vienna -- Vienne -- Beć, ...; Brüssel -- Brussels -- Bruxelles, ...). In bibliographic entries, the publisher location normally should be given in the main language of the document.

So I wonder whether `\DeclareListFormat{location}` should not use
```tex
\ifbibstring{#1}{\bibstring{#1}}{#1\isdot}%
```
(actually just like patent location). Biblatex could include bibkeys for the most common publisher locations, but even if it didn't, it would help to maintain an own list. Currently I am working around this with `@string`s which change depending on the document language.

Another related issue which I am not sure how to deal with concerns the (mainly US?) convention to give state names (Cambridge, MA), the use and form (cf. Cambridge, Mass.) of which seems to depend on style. Maybe this could be approached via data annotation, or `\locstate{<bibkey>}` macros. Anyways, this seems less central than the issue above.

plk:
I would support this - I always wondered about the `location` and localisation - @moewew?

jspitz:
If you consider to provide a selected list of some common locations, from the top of my head with a brief glance at [this Wikipedia list](https://en.wikipedia.org/wiki/List_of_names_of_European_cities_in_different_languages)

- Antwerp
- Athens
- Basle
- Berne
- Beijing
- Belgrade
- Bolzano
- Brescia
- Brussels
- Cape Town
- Cairo
- Cologne
- Copenhagen
- Florence
- Geneva
- Gothenbourg
- The Hague
- Hanover
- Leuwen
- Lisbon
- Lucerne
- London
- Milan
- Moscow
- Munich
- Nijmegen
- Nuremberg
- Rome
- Saint Petersburg
- Strasbourg
- Tokyo
- Trieste
- Venice
- Vienna
- Zurich

Maybe also:

- Frankford (Frankfurt/M.)

moewew:
Hmmm. Thing is, this will probably be a huge burden to maintain and we probably have to add all sorts of cities once people start asking for it.  (At some point we might even run into performance issues if we take it too far.) Do we really want this?

plk:
We could do a short list - any others need to be added by users - the above list isn't too long and I doubt there would be performance issues with that.

pauloney:
My understanding of the objective might not be complete, but I imagine --
[...]

PR #1403, opened 2024-12-07
Make location a key list (core of #1390)

jspitz:
No bibkeys provided yet.


jspitz:
I was wondering whether we shouldn't do the same with `venue`. But I didn't find where it is defined.

jspitz:
> I was wondering whether we shouldn't do the same with `venue`. But I didn't find where it is defined.

`venue` is now considered as well

jspitz:
BTW an argument pro making `venue` localizable: at least by default, `evendate` is localized, so you might end up with phrases like `Vienna, 25. Dezember 2024` or `Wien, December 25, 2024`, which looks really odd.

PR #1421, opened 2025-04-14
fixing month name

mafsi:
— in situations like `vol. II, nr. 18 (20 Septembrie 1972)` we need to put month in lowercase 
— the same in situation above applies in  `accesat 13 Martie 2023`


moewew:
Thank you very much. Merged.

Issue #699, opened 2018-01-29
arXiv links use http (not https)


Socob:
It would be nice if the arXiv links generated by biblatex could use the https protocol (instead of http). As far as I can tell, arxiv.org redirects any http traffic to https anyway. The relevant line in the code seems to be this one:

https://github.com/plk/biblatex/blob/5936adba211ab1e547ac5fe59e0836ca7ff6b316/tex/latex/biblatex/biblatex.def#L469

moewew:
Thank you for the request.

I have pushed a commit to that effect.

moewew:
Done in `biblatex` 3.11.

PR #1460, opened 2026-02-08
Use HTTPS for HDL links

sim642:
This is analogous to #699.


moewew:
Thank you very much. Merged.

PR #1469, opened 2026-03-28
Fix basque language date output

Josu-A:
The current implementation of dates for the basque language isn't correct. This PR implements the dates according to the [37th rule of Euskaltzaindia](https://www.euskaltzaindia.eus/dok/arauak/Araua_0037.pdf) and uses the [basquedate package](https://ctan.org/pkg/basque-date) as a reference.

Feel free to comment on anything that seems wrong.

### DIFF
--- a/tex/latex/biblatex/lbx/basque.lbx
+++ b/tex/latex/biblatex/lbx/basque.lbx
@@ -9,41 +9,86 @@
   \let\finalandcomma=\empty
   \let\finalandsemicolon=\empty
   \def\mkbibordinal{\mkbibmascord}%
-  \providecommand*{\sptext}{\textsuperscript}% \sptext -> spanish.ldf (babel)
-  %\protected\def\mkbibmascord#1{\stripzeros{#1}\sptext{o}}%
+  \providecommand*{\sptext}{\textsuperscript}%
   \protected\def\mkbibmascord#1{\stripzeros{#1}\adddot}%
-  %\protected\def\mkbibfemord#1{\stripzeros{#1}\sptext{a}}%
   \protected\def\mkbibfemord#1{\stripzeros{#1}\adddot}%
   \protected\def\mkbibneutord{\mkbibmascord}%
+  \def\blx@basque@mon@i{urtarrila}%
+  \def\blx@basque@mon@ii{otsaila}%
+  \def\blx@basque@mon@iii{martxoa}%
+  \def\blx@basque@mon@iv{apirila}%
+  \def\blx@basque@mon@v{maiatza}%
+  \def\blx@basque@mon@vi{ekaina}%
+  \def\blx@basque@mon@vii{uztaila}%
+  \def\blx@basque@mon@viii{abuztua}%
+  \def\blx@basque@mon@ix{iraila}%
+  \def\blx@basque@mon@x{urria}%
+  \def\blx@basque@mon@xi{azaroa}%
+  \def\blx@basque@mon@xii{abendua}%
   \protected\def\mkbibdatelong#1#2#3{%
-    \iffieldundef{#3}
-      {}
-      {\stripzeros{\thefield{#3}}%
-       \iffieldundef{#2}{}{\nobreakspace de\space}}%
-    \iffieldundef{#2}
-      {}
-      {\mkbibmonth{\thefield{#2}}%
-       \iffieldundef{#1}{}{\nobreakspace de\space}}%
-    \iffieldbibstring{#1}
-      {\bibstring{\thefield{#1}}}
-      {\dateeraprintpre{#1}\mkyearzeros{\thefield{#1}}}}%
-  \protected\def\mkbibdateshort#1#2#3{%
     \iffieldundef{#1}
       {}
-      {\mkdayzeros{\thefield{#1}}%
-       \iffieldundef{#2}{}{\bibdatesep}}%
+      {\dateeraprintpre{#1}\mkyearzeros{\thefield{#1}}%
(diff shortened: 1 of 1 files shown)


moewew:
Thanks for reporting this and providing a fix as well!

The 100-way case distinction for "eko/ko" looks a bit heavy, is there some sort of pattern we can use here? It appears to me that if the number is an odd multiple of 10, one plus an even multiple of 10 or ends in 5, we have eko and otherwise ko. Is that correct?

And we need "a" for all numbers except 11?

What about the declined month names? Do we ever need to use "urtarrilaren" vs "urtarrila" in `biblatex`?

Josu-A:
> The 100-way case distinction for "eko/ko" looks a bit heavy, is there some sort of pattern we can use here? It appears to me that if the number is an odd multiple of 10, one plus an even multiple of 10 or ends in 5, we have eko and otherwise ko. Is that correct?

This would be somewhat correct yes, you made me realise there's a mistake on the logic used. In basque the number 100 (and 200, 300, ... 900) ends with the letter "n" (ehun, berrehun, hiruehun, hirurehun, ..., bederatziehun) and therefore needs "-eko", while the number 0 ends with "o" (zero) and needs "-ko". Other magnitudes such as 1,000 (mila), 1,000,000 (milioi), 1,000,000,000 (mila milioi), 1,000,000,000,000 (bilioi), etc. also end with "-ko", so there's only an edge case when the number ends with 100, 200, ..., and 900.

I've put up a possible implementation here, but it has issues. Calculating the year digits (`yearlasttwo` / `yearmod`) returns undesirable numbers. E.g. starting with initial year
  - `1999`:  makes `yearlasttwo = -1` and `yearmod = -1 `, when I would expect it to be `99 | 19`.
  - `2001`-`2009`:  makes `yearlasttwo = 1`-`9` and `yearmod = 1`-`9`, hurray.
  - `2010`-`2019`: makes `yearlasttwo = 10`-`19` and `yearmod = -10`-`-1`, when I would expect them to be `10 | 10` - `19 | 19`.
  - `2020`-`2029`:  makes `yearlasttwo = 20`-`29` and `yearmod = 0`-`9`, hurray.
  - `2030`-`2039`:  makes `yearlasttwo = 30`-`39` and `yearmod = -10`-`-1`, when I would expect them to be `30 | 10` - `39 | 19`.
  - etc.

There's something about latex division that I don't understand, but if this is fixed it should mean all edge cases are covered.

```tex
    \iffieldundef{#1}
      {}
      {\dateeraprintpre{#1}\mkyearzeros{\thefield{#1}}%
       \ifboolexpr{not test {\iffieldundef{#2}} or not test {\iffieldundef{#3}}}
         {\ifnumequal{\thefield{#1}}{0}%
          {ko}% zero (0)
          {\edef\lbx@basque@yearlasttwo{\the\numexpr\thefield{#1}-(\thefield{#1}/100)*100\relax}%
           \ifnumequal{\lbx@basque@yearlasttwo}{0}
            {\ifnumequal{\the\numexpr\thefield{#1}-(\thefield{#1}/1000)*1000\relax}{0}
             {ko}% mila, bi mila, milioi, ... (1000, 2000, 1 000 000, ...)
             {eko}}% ehun, berrehun, ..., bederatziehun (100, 200, ..., 900)
            {\edef\lbx@basque@yearmod{\the\numexpr%
              \lbx@basque@yearlasttwo-((\lbx@basque@yearlasttwo)/20)*20\relax}%
             \ifboolexpr{
               test {\ifnumequal{\lbx@basque@yearmod}{1}}% 1, 21, 41, 61, 81
               or test {\ifnumequal{\lbx@basque@yearmod}{5}}% 5, 25, 45, 65, 85
               or test {\ifnumequal{\lbx@basque@yearmod}{10}}% 10, 30, 50, 70, 90
               or test {\ifnumequal{\lbx@basque@yearmod}{15}}% 15, 35, 55, 75, 95
             }
               {eko}
               {ko}
            }}}%
      {}}%
```

> And we need "a" for all numbers except 11?

Basque has two forms to write date, for example:

- 1983ko martxoaren 7a
- 1983ko martxoak 7

The first one (longer one) is currently the most used one, where the definite article "a" is added to the day of the month. When the pronounced number ends with an "a" (rare), we drop the definite article to phonetically smooth it out; with dates this happens with the numbers 11 and 31 [[1]](https://www.euskaltzaindia.eus/index.php?option=com_ebe&Itemid=1161&task=bilaketa&id=388).

> What about the declined month names? Do we ever need to use "urtarrilaren" vs "urtarrila" in `biblatex`?

If you add a bib entry using `date` (e.g. `date = {1983-03-07}`), the desired full length output would be "1983ko martxoaren 7a". But what if you lacked the day (or even month)? You would have to use the fields `year` and `month`. Without having the "martxoa" base word the output you would get is "1983ko martxoaren", which is wrong because the genitive form "-ren" expects something else to appear after it. Therefore, the correct form would be "1983ko martxoa".

moewew:
Alright, I implemented this in https://github.com/moewew/biblatex/commit/18a35479babfe67aced5795f29e9e8e2c3334995, I just switched around the two versions of the month strings. To me it seemed that the genitive form is somehow more "special" so those forms new bibstrings. Let me know if that is somehow wrong or confusing. If you say that this looks good I will merge the PR manually.

Josu-A:
Seems great. The division inconsistencies of my implementation also appear to be fixed after testing a few range of years. Thanks for your time!

moewew:
Excellent! Thanks for the feedback. Merged manually in https://github.com/plk/biblatex/commit/a29995037732e15e44dd893767631692fe59a4f6.

PR #1471, opened 2026-04-11
Fixes #1470 and disables the footnote-patch-failed warning in this case

### LINKED ISSUE #1470: Footnote detection fails if \DocumentMetadata is used
Ref.: https://tex.stackexchange.com/q/761788/

```latex
\DocumentMetadata{lang=en-GB,}

\documentclass{article}
\usepackage[style=authoryear-icomp]{biblatex}
\bibliography{biblatex-examples}
\begin{document}

Example\footnote{%
  \Autocite{aksin}.%
}.

Example\footnote{%
  \Autocite{aksin}.%
}.

Example\footnote{%
  Feetnotes are not a thing.%
}.

Example\footnote{%
  Feetnotes are not a thing.%
}.

Example\footnote{%
  Feetnotes are not a thing.%
}.

Example\footnote{%
  \Autocite{aksin}.%
}.
\end{document}
```
Output:
```
1(Aksın et al. 2006).
2 (Ibid.).
3Feetnotes are not a thing.
4 Feetnotes are not a thing.
5 Feetnotes are not a thing.
6 (Ibid.).
```
rather than the expected
```
1(Aksın et al. 2006).
2 (Ibid.).
3Feetnotes are not a thing.
4 Feetnotes are not a thing.
5 Feetnotes are not a thing.
6 (Aksın et al. 2006).
```
In some cases (e.g. if `setspace` is loaded before `biblatex`) a warning is issued. However, in most cases, footnote detection silently fails.

cfr42:
Fixes footnote detection if `\DocumentMetadata` is used and disables the footnote-patch-failed warning in this case.

Fixes https://github.com/plk/biblatex/issues/1470

The reason no warning is issued except by chance is that Biblatex's patch succeeds, but is then silently clobbered by `latex-lab` in a later hook. To avoid this, the code needs to be delayed.

This does not use any LaTeX internals, since sufficient public hooks are now available to inject footnote detection.

This is **extremely** minimally tested.

### DIFF
--- a/tex/latex/biblatex/biblatex.sty
+++ b/tex/latex/biblatex/biblatex.sty
@@ -419,6 +419,13 @@
        {}
        {}}
     {}%
+  \IfDocumentMetadataT{% will only work for classes which use the hook
+    \AddToHook{begindocument}{%
+      \AddToHook{fntext/begin}{\toggletrue{blx@footnote}}%
+      \AddToHook{fntext/end}{\togglefalse{blx@footnote}}%
+    }%
+    \togglefalse{blx@tempa}%
+  }%
   \iftoggle{blx@tempa}
     {\blx@warning@noline{%
        Patching footnotes failed.\MessageBreak


u-fischer (commented):
I didn't test or check all the variants, but it would probably be saner (and faster) to disable the patching of footnotes completely if `\DocumentMetadata` is detected.

moewew:
Thanks for reporting this and suggesting the fix!

@u-fischer Is there grouping in `fntext` that we can rely so that we can get away with setting only `\AddToHook{fntext/begin}{\toggletrue{blx@footnote}}` or do we also need the off switch `\AddToHook{fntext/end}{\togglefalse{blx@footnote}}`? In my tests, things worked without the `fntext/end` hook, but this may have been accidental.

u-fischer:
@moewew  there is a `\color@begingroup`/`\color@endgroup`  around the text and the hook, so it is always grouped.

moewew:
Should be fixed in https://github.com/plk/biblatex/commit/3554a1f5340869ff38eacc3240ce8846fba42c4d, where we no longer try to patch footnote commands with `\DocumentMetadata`.

Issue #1477, opened 2026-05-21
Feature request: Add `category` option to `\addbibresource`


nilsjor:
In the case where bibliographies are split into different files, it would be useful to be able to assign a `category` to _all_ entries in that file when loaded, by using the following suggested additional option:

```tex
\DeclareBibliographyCategory{primary}
\DeclareBibliographyCategory{secondary}

\addbibresource[category = primary]{...}
\addbibresource[category = secondary]{...}
```

instead of 

```tex
\DeclareBibliographyCategory{primary}
\DeclareBibliographyCategory{secondary}

\addbibresource{...}

\addtocategory{primary}{
  keyP1, 
  keyP2, 
  keyP3, 
  ...
}

\addtocategory{secondary}{
  keyS1, 
  keyS2, 
  keyS3, 
  ...
}
```

For completeness, it might be sensible to also allow more than one category, using the appropriate syntax; `category = {cat1, cat2, ...}` or whatever would be correct here. The desired behavior would then be that _all_ provided categories are applied to the entries in that file

moewew:
Categories would be a bit harder to do, since they essentially live in the `.tex` file, but the entry data passed to the `.tex` file does not remember which `.bib` file it came from. You can get `keywords`, though, which often offer the same functionality.

See for example https://tex.stackexchange.com/q/35279/35864

```latex
\documentclass{article}
\usepackage[utf8]{inputenc}
\usepackage[backend=biber,style=authoryear]{biblatex}
\usepackage{csquotes}

\DeclareSourcemap{
  \maps[datatype=bibtex, overwrite]{
    \map{
      \perdatasource{\jobname-primary.bib}
      \step[fieldset=keywords, fieldvalue=primary, append]
    }
    \map{
      \perdatasource{\jobname-secondary.bib}
      \step[fieldset=keywords, fieldvalue=secondary, append]
    }
  }
}

\begin{filecontents}{\jobname-primary.bib}
@BOOK{hectic,
  AUTHOR    = {Henry Hectic},
  TITLE     = {How Horticulturalists Howl},
  PUBLISHER = {Honorary Books: Henage},
  YEAR      = {2000}
}
\end{filecontents}
\begin{filecontents}{\jobname-secondary.bib}
@BOOK{flutter,
  AUTHOR    = {Frederick Flutter},
  TITLE     = {Fraternising with Flowers},
  PUBLISHER = {Frippery Pamphlets: Folkestone},
  YEAR      = {1995}
}
[...]

PR #1480, opened 2026-05-29
Add `category=` option to `\addbibresource` and related commands (#1477)

nilsjor:
This PR implements the `category=` option for `\addbibresource`, `\addglobalbib`, and `\addsectionbib`, allowing all entries from a given resource to be automatically assigned to one or more bibliography categories without manually enumerating entries with `\addtocategory`. (Issue #1477)

### Testing

Regression tests added via the standard `l3build` test framework. Existing test suite remains clean (`l3build check` passes unchanged).

### Usage

```latex
% Single category
\addbibresource[category=primary]{sources-primary.bib}

% Multiple categories (comma-separated)
\addbibresource[category={primary,secondary}]{sources-both.bib}

% Works with other options too
\addbibresource[category=primary, label=prim]{sources-primary.bib}

% Also available on \addglobalbib and \addsectionbib
\addglobalbib[category=primary]{global.bib}
\addsectionbib[category=secondary]{section.bib}
```

The category must still be declared with `\DeclareBibliographyCategory` beforehand. This mirrors the existing behaviour of `\addtocategory` in that respect.

### Architecture

The implementation works through biber's sourcemap layer rather than by tracking entries at the LaTeX level. For each resource that carries a `category=` option, a user-level sourcemap is written to the BCF that appends a sentinel keyword tag `blx@srccat@<name>` to every entry originating from that datasource. A hook on `blx@hook@bblitem@*` then checks for the tag with `\ifkeyword` at BBL-read time and calls `\blx@addtocategory`, replicating the effect of `\addtocategory` without requiring the user to enumerate entries.

This intermediate-keyword approach was suggested by @moewew in the issue thread — thanks for pointing in that direction, it turned out to be exactly the right hook to reach for.

When multiple categories are specified for one resource, a single `<maps>` block is emitted with a comma-separated keyword string (`blx@srccat@primary,blx@srccat@secondary`), which biber's `\keyw{}` field handles correctly.

The category processing is centralised in `\blx@addbib@applycats`, called from `\blx@addbib@register`. Since `\blx@addbib@resource` calls `\blx@addbib@register`, and `\blx@addbib@global` calls `\blx@addbib@resource`, all three user-facing commands share one code path with no duplication.

### Tests added

| Test | What it covers |
|---|---|
| `issue-1477` | Single category per resource, two resources |
| `issue-1477-baseline` | Plain `\addbibresource` with no `category=`, guards against regressions |
| `issue-1477-multiple` | One resource assigned to two categories via `category={primary,secondary}` |
| `issue-1477-opts` | `category=` combined with `label=` and `type=`, verifying option parsing robustness |
| `issue-1477-globalbib` | `category=` on `\addglobalbib` |
| `issue-1477-sectionbib` | `category=` on `\addsectionbib` with resources supplied via `\refsection` |
| `issue-1477-error` | Correctly throwing an undeclared category error if used without `\DeclareBibliographyCategory ` |

### A note on tooling

I used GitHub Copilot (Claude Sonnet 4.7) as a development aid while working on this patch. Since I am not an experienced LaTeX/biblatex developer, I took extra care to manually review, refine, and regression-test all suggested changes using the project's `l3build` test framework.

I also intentionally left the commit history unsquashed so the individual changes and refinements can be reviewed/audited step-by-step during code review rather than appearing as a single opaque patch.

I am very happy to revise the implementation during review and I would genuinely appreciate feedback on code style, architecture, or better idiomatic approaches within biblatex/biber. I expect there are parts that can be improved, and I look forward to cooperating on that.


moewew:
Thank you for contributing this. I've had quite some time to think about it (sorry it took so long) and I have decided that I don't want to merge this pull request.

For starters I'm not sure if I think that the implementation via keywords that I suggested as an ad-hoc solution is the best way for an "official implementation" of this feature. It feels a little dishonest to sneak an internal keyword into the mix that could be exposed accidentally (granted, very unlikely if you choose an internal prefix, but still).

Secondly, the code shown in https://github.com/plk/biblatex/issues/1477#issuecomment-4547921676 demonstrates that it is not actually too complicated for users to add some code to their preamble if they want something like this. If they have to do that explicitly, they are also in a position to have an eye on the side-effects of this implementation.

Then there is the point that this code is mainly AI generated. I'm generally impressed by the quality, but there are some quirks in the code that I don't quite understand (for example I really don't understand why 
https://github.com/plk/biblatex/pull/1480/changes#diff-6c60f2550cf6c831795d89e5f8e914df013a7d66374d44622bb9e872f591f8f3R11396-R11399 should be necessary). And it is fairly low-level code at that (so always a little more tricky to review). I think for now I'd rather not add code that is mainly AI generated and would prefer contribution where a human can justify the coding choices that were made (so we can ask them) and does not "just" vouch for testing the output (which is very important - don't get me wrong).

PR #1494, opened 2026-09-05
Add Azerbaijani localization

abdanar:
This PR adds Azerbaijani localization support to `biblatex`.

It includes:

* a new `azerbaijani.lbx` localization file;
* registration of Azerbaijani language strings;
* Azerbaijani-aware case mapping;
* localization of bibliography strings, contributor roles, language names, dates, pagination, publication states, backreferences, and related entries;
* Azerbaijani-specific formatting where the standard English word order does not produce natural output;
* documentation and changelog updates.

The Azerbaijani-specific related-entry and shorthand-introduction handling is kept local to the language and preserves the standard `biblatex` machinery, including explicit user overrides such as `relatedstring` and `shorthandintro`.

The localization was tested with LuaLaTeX and Biber across ordinary bibliography entries, contributor-role combinations, language-name capitalization, related entries, URL access dates, shorthand introductions, backreferences, and abbreviated author lists. Equivalent English fixtures were also compiled to check for regressions.


moewew:
Thank you very much for contributing a new localization.

Unfortunately, I didn't yet have time to review the entire file. I'd like to have a very close look at the `related` changes before we merge. Maybe there is a general principle that we can abstract here for other/future loclizations.

